Issue #89, opened 2018-09-11
Serializing absent optional as empty body should omit content-type


dansanduleac:
Identified in https://github.com/palantir/conjure-verification/issues/45

The [wire spec](https://github.com/palantir/conjure/blob/develop/docs/spec/wire.md#http-requests) states that all endpoints defining a body parameter must set one of two content-type headers:

> Content-Type header - For Conjure endpoints that define a body argument, a Content-Type header MUST be added. If the body is of type binary, the content-type `application/octet-stream` MUST be used. Otherwise, clients MUST send `Content-Type: application/json`.

However, a bit further up under "Body parameter" it suggests (even recommends) that clients should send an empty body in the case of an absent optional body parameter, which [doesn't seem to be valid JSON](https://stackoverflow.com/a/40624629/274699):

> [when] the de-aliased argument is type `optional<T>` and the value is not present: it is RECOMMENDED to send an empty request body, although clients MAY alternatively send the JSON value `null`

Since we probably have to support this use case, we should say that content-type MUST NOT be set when sending an absent optional body as an empty body.

iamdanfox:
One further argument in support of this wire-spec change is that Chrome actually strips the `Content-Type` header if the body is empty!

cakofony:
If there's no content it cannot possibly have a type.

Except... if we send a binary request, binary data is allowed to be zero-bytes, and we have no way to differentiate between absent and present but zero bytes where responses can specify 200/204 status, which itself isn't quite right because no content should always use the 204 status...  Additional musings in #186

PR #93, opened 2018-09-21
[fix] update spec regarding content-type and 204

qinfchen:
fixes https://github.com/palantir/conjure/issues/89

## Before this PR
Setting `Content-type: application/json` header for empty body requests causes serialization errors because empty body is not a valid json blob.

## After this PR
the spec now explicitly states that `content-type` header cannot be set to `application/json` on empty body requests


iamdanfox:
Closing out due to conflicts

PR #123, opened 2018-10-24
[improvement] Add 'deprecated' support to field and enum values to Conjure

nathklei:
## Before this PR
If an application developer wished to deprecate a field or enum value in his or her model, they had no way to communicate that in their conjure definition.

## After this PR
Application developers may now specify fields and enum values that are deprecated, along with documentation as to why and what the replacement is, if any. This indicates to consumers and tooling that these fields and enum values are subject for removal or should be removed during a cleanup sweep.


iamdanfox:
Sure thing! I have a sketch of how IR change be changed that I think will be turned into an RFC, will make sure to cc you guys!

davidscottcohen:
Any update here?  I'd love to have this in the next 1-2 months.

iamdanfox:
released in 4.8.0, we actually did this without an IR revision to make life simpler: https://github.com/palantir/conjure/releases/tag/4.8.0

PR #169, opened 2018-11-25
[fix] Fix getting started build.gradle

ferozco:
## Before this PR
Users exactly following the getting started guide would be unable to resolve all their build script dependencies and would get a very confusing error.

## After this PR
Users following the getting started guide should no longer encounter any issues.

Thanks to MrFantastik from Lobster for [point this out](https://lobste.rs/s/egbgwb/introducing_conjure_palantir_s#c_vzpbkj)


ferozco:
👍

PR #180, opened 2018-12-04
[improvement] Allow dots in endpoint paths

### LINKED ISSUE #177: Support dots in path
## Motivation

Having a versioned path such as `/some/v1.2/path` breaks the IR compilation:

```
Segment v1.2 of path /some/v1.2/path did not match required segment patterns ^[a-zA-Z][a-zA-Z0-9_-]*$ or parameter name patterns ^\{[a-z][a-z0-9]*([A-Z0-9][a-z0-9]+)*}$ or ^\{[a-z][a-z0-9]*([A-Z0-9][a-z0-9]+)*(\Q:.+\E|\Q:.*\E)}$
```

This is a common pattern we use to maintain endpoints between microservices and between service and frontend clients.

## Proposal

If there's no specific reason that `.`s are not supported in paths, perhaps the `PathString .SEGMENT_PATTERN` could be modified to accept `.`s.

Thanks guys for open sourcing this! Really missed having this library around, i remember it being super helpful for reducing boilerplate

shzhng:
## Before this PR
IR compilation would break if a service endpoint have a path definition of `some/v1.2/path`

## After this PR
Allows `.`s in the path, fixes #177 

@iamdanfox

### DIFF
--- a/conjure-core/src/main/java/com/palantir/conjure/defs/validator/HttpPathValidator.java
+++ b/conjure-core/src/main/java/com/palantir/conjure/defs/validator/HttpPathValidator.java
@@ -36,7 +36,7 @@ public final class HttpPathValidator {
 
     private HttpPathValidator() {}
 
-    public static final String PATTERN = "[a-z][a-z0-9]*([A-Z0-9][a-z0-9].+)*";
+    public static final String PATTERN = "[a-z][a-z0-9]*([A-Z0-9][a-z0-9]+)*";
     private static final Pattern SEGMENT_PATTERN = Pattern.compile("^[a-zA-Z][a-zA-Z0-9._-]*$");
     private static final Pattern PARAM_SEGMENT_PATTERN = Pattern.compile("^\\{" + PATTERN + "}$");
     private static final Pattern PARAM_REGEX_SEGMENT_PATTERN =
--- a/conjure-core/src/main/java/com/palantir/conjure/defs/validator/HttpPathValidator.java
+++ b/conjure-core/src/main/java/com/palantir/conjure/defs/validator/HttpPathValidator.java
@@ -36,8 +36,8 @@ public final class HttpPathValidator {
 
     private HttpPathValidator() {}
 
-    public static final String PATTERN = "[a-z][a-z0-9]*([A-Z0-9][a-z0-9]+)*";
-    private static final Pattern SEGMENT_PATTERN = Pattern.compile("^[a-zA-Z][a-zA-Z0-9_-]*$");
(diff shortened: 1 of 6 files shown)


palantirtech:
Thanks for your interest in [palantir/conjure](https://github.com/palantir/conjure), @shzhng! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

conjure-core/src/main/java/com/palantir/conjure/defs/validator/HttpPathValidator.java
```diff
@@ -37,7 +37,7 @@
     private HttpPathValidator() {}
 
     public static final String PATTERN = "[a-z][a-z0-9]*([A-Z0-9][a-z0-9]+)*";
-    private static final Pattern SEGMENT_PATTERN = Pattern.compile("^[a-zA-Z][a-zA-Z0-9_-]*$");
+    private static final Pattern SEGMENT_PATTERN = Pattern.compile("^[a-zA-Z][a-zA-Z0-9._-]*$");
```
shzhng: @iamdanfox slightly confused as i see this block duplicated in both this file as well as com/palantir/conjure/parser/services/PathString.java

should i be making the change to both?

conjure-core/src/main/java/com/palantir/conjure/parser/services/PathString.java
```diff
@@ -39,7 +39,7 @@
     /** Returns the well-formed path associated with this path definition. */
     public abstract Path path();
 
-    private static final Pattern SEGMENT_PATTERN = Pattern.compile("^[a-zA-Z][a-zA-Z0-9_-]*$");
+    private static final Pattern SEGMENT_PATTERN = Pattern.compile("^[a-zA-Z][a-zA-Z0-9._-]*$");
```
shzhng: after some digging, i presume this when actual parsing happens in preparation for IR compilation

assuming this was copy pasta, i can also refactor these into a single place. let me know if my assumption is wrong before i proceed

ferozco: It seems like we're doing identical validation. I think for consistency we should remove the parse time check and do it as part of the httpPathValidator

ferozco:
@shzhng can you dedupe the validation so that we can get this in?

shzhng:
@ferozco sure thing, i'll try to get around to this sometime today

ferozco:
👍

PR #191, opened 2019-01-09
Clarification: Allow lists and sets of query parameters.

cakofony:


docs/spec/conjure_definitions.md
```diff
@@ -355,7 +355,7 @@ An object representing an argument to an endpoint.
 
 Field | Type | Description
 ---|:---:|---
-type | [ConjureType][] | **REQUIRED**. The type of the value of the argument. The type name MUST exist within the Conjure definition. If this ArgumentDefinition has a param-type of `body` then there are no restrictions on the type. If the param-type is `path`, `header` or `query` then the de-aliased type MUST be an enum, a primitive or an optional of one of these. Additionally, the type MUST NOT be `bearertoken` when the param-type is `header` or `query`.
+type | [ConjureType][] | **REQUIRED**. The type of the value of the argument. The type name MUST exist within the Conjure definition. If this ArgumentDefinition has a param-type of `body` then there are no restrictions on the type. If the param-type is `path`, `header` or `query` then the de-aliased type MUST be an enum, a primitive or an optional of one of these. `query` type parameters may accept lists and sets in addition to optional containers. Additionally, the type MUST NOT be `bearertoken` when the param-type is `header` or `query`.
```
cakofony: This is difficult to read. I added:
```
`query` type parameters may accept lists and sets in addition to optional containers.
```

markelliot:
Why do we want this?

cakofony:
This is a clarification for functionality that is already heavily relied upon. I can send internal links to a few places if you like.

markelliot:
I’d like to see references. What is the serialization format?


________________________________

cakofony:
Standard http:

```yaml
ids:
  param-type: query
  param-id: id
  param-type: string
```

results in `/api?id=1&id=2&id=3`

iamdanfox:
From Carter's list, I can see five foundry products that use this.  It is implemented in the conjure-typescript-client and is in active use from our notifications frontend.

https://github.com/palantir/conjure-typescript-runtime/blob/develop/packages/conjure-client/src/fetchBridge/fetchBridge.ts#L181-L187

Looks like it was actually just a mistake that we didn't include this in the spec right from the start. Can we also make sure to mention this in the wire format doc?: `?key=value1&key=value2&key=value3`

cakofony:
> Can we also make sure to mention this in the wire format doc?

Updated :-)

docs/spec/conjure_definitions.md
```diff
@@ -355,7 +355,7 @@ An object representing an argument to an endpoint.
 
 Field | Type | Description
 ---|:---:|---
-type | [ConjureType][] | **REQUIRED**. The type of the value of the argument. The type name MUST exist within the Conjure definition. If this ArgumentDefinition has a param-type of `body` then there are no restrictions on the type. If the param-type is `path`, `header` or `query` then the de-aliased type MUST be an enum, a primitive or an optional of one of these. Additionally, the type MUST NOT be `bearertoken` when the param-type is `header` or `query`.
+type | [ConjureType][] | **REQUIRED**. The type of the value of the argument. The type name MUST exist within the Conjure definition. If this ArgumentDefinition has a param-type of `body` then there are no restrictions on the type. If the param-type is `path`, `header` or `query` then the de-aliased type MUST be an enum, a primitive or an optional of one of these. `query` type parameters may accept lists and sets in addition to optional containers, which are represented by repeating the `param-id`, for example `?key=value1&key=value2&key=value3`. Additionally, the type MUST NOT be `bearertoken` when the param-type is `header` or `query`.
```
iamdanfox: @cakofony I think I'd prefer to keep the actual examples in `wire.md` to minimize any distractions in this one.

cakofony: Makes sense, sorry!

iamdanfox:
This now also matches the actual validation we have in Conjure!

https://github.com/palantir/conjure/blob/676856d89e073f02490e45eb71fe21da96fc2023/conjure-core/src/main/java/com/palantir/conjure/defs/validator/EndpointDefinitionValidator.java#L233-L234

cakofony:
👍

PR #201, opened 2019-01-18
Mention conjure-rust on the README

iamdanfox:
cc @sfackler


ferozco:
👍

sfackler:
Seems like we should probably split the generator out from the runtime libraries like the other implementations do?

PR #207, opened 2019-01-24
[fix] Prevent optional<binary> serialization in HTTP body

uschi2000:


uschi2000:
fyi @cakofony @iamdanfox

docs/spec/wire.md
```diff
@@ -193,8 +193,9 @@ Further non-successful status codes are defined in the Conjure errors section be
 ### 3.2. Response body
 Conjure servers must serialize return values using the [JSON format][] defined below, unless:
 
-  - the de-aliased return type is `optional<T>` and the value is not present: servers must omit the HTTP body.
-  - the de-aliased return type is `binary`: servers must write the raw bytes as an octet-stream.
+  - the de-aliased return type is `optional<T>` and the value is not present: servers must return an empty HTTP body
```
cakofony: I think we should specify 204 response code, and no http response body

docs/spec/wire.md
```diff
@@ -193,8 +193,9 @@ Further non-successful status codes are defined in the Conjure errors section be
...
+  - the de-aliased return type is `optional<T>` and the value is present: servers must write the raw bytes as an octet-stream
```
cakofony: Did you mean optional\<binary> rather than optional\<T>?

docs/spec/wire.md
```diff
@@ -193,8 +193,9 @@ Further non-successful status codes are defined in the Conjure errors section be
...
+  - the de-aliased return type is `binary`: servers must write the raw bytes as an octet-stream
```
cakofony: Perhaps we should reword to specify that any empty optional regardless of the contained type should result in a 204 and no body, and that we expect identical responses for T and optional.present(T) for all values of T.

docs/spec/wire.md
uschi2000: the 204 thing is defined above. "empty body" is equivalent to "no body", right?

docs/spec/wire.md
uschi2000: yep, fixed

docs/spec/wire.md
uschi2000: can re-arrange in a follow-up, but as it stands the structure asks me to just talk about the body in this section

docs/spec/wire.md
cakofony: It should be, though an empty body could be construed to have a content type, where no body would not, and aligns with our “no content” response code.

I’m happy either way

uschi2000:
@iamdanfox thoughts on this?

uschi2000:
@bmoylan I would happily do the enforcement, but only after we agree this is something we should and can do. @iamdanfox ?

docs/spec/conjure_definitions.md
```diff
@@ -361,6 +361,8 @@ deprecated | `string` | Documentation for why this argument is deprecated. [Comm
 param&#8209;id | `string` | An identifier to use as a parameter value. If the param type is `header` or `query`, this field may be populated to define the identifier that is used over the wire. If this field is undefined for the `header` or `query` param types, the argument name is used as the wire identifier. Population of this field is invalid if the param type is not `header` or `query`.
 param&#8209;type | [ArgumentDefinition.ParamType][] | The type of the endpoint parameter. If omitted the default type is `auto`.
 
+Arguments with parameter type `body` MUST NOT be of type `optional<binary>`, or, intuitively, of a type that reduces to `optional<binary>` via unfolding of alias definitions and nested `optional`.
```
iamdanfox: :+1: definitely in sync that we want to ban this.  It may be defining the 'de-alias' function at the top of this file so we don't have to hand-wave about unfolding aliases?  Also optional<optional<T>> is banned already, so don't need to mention that here.

uschi2000:
Added a validator against `optional<binary>`, @iamdanfox

conjure-core/src/main/java/com/palantir/conjure/defs/validator/EndpointDefinitionValidator.java
```diff
@@ -136,14 +138,35 @@ public void validate(EndpointDefinition definition) {
...
 
+    @com.google.errorprone.annotations.Immutable
+    private static final class NoOptionalBinaryBodyParamValidator implements ConjureValidator<EndpointDefinition> {
+        @Override
+        public void validate(EndpointDefinition definition) {
+            boolean hasOptionalBinaryBody = definition.getArgs()
+                    .stream()
+                    .filter(entry -> entry.getParamType().accept(ParameterTypeVisitor.IS_BODY))
+                    .map(ArgumentDefinition::getType)
+                    .filter(type -> type.accept(TypeVisitor.IS_OPTIONAL))
```
ferozco: You'll want to use a dealiasingTypeVisitor, like we do in other validators, to protect against alias of `optional<binary>`

uschi2000: like in NoComplexPathParamValidator, ja?

uschi2000: done

cakofony:
Thanks all!

PR #231, opened 2019-02-18
[improvement] Simplify Case conversion logic

ellisjoe:
## Before this PR
Bonus logic for each `Case` type.

## After this PR
Each `Case` corresponds to a `CaseFormat` that can be used in a generic way to convert between the types.


dansanduleac:
Wow can't believe we had this duplication in there...

PR #233, opened 2019-02-18
Unify regex's

aborochoff:
There should only be one source of the regex's?

## Before this PR

Regex's for field names are not the same.

## After this PR

Regex's for field name validation are now the same.  Going to actually take a look at why the patterns are in 2 different places...


aborochoff:
https://github.com/palantir/conjure/pull/234

PR #234, opened 2019-02-18
[fix] Get rid of the patterns in FieldName to allow single character field names

aborochoff:
## Before this PR

There were 2 validations for field names.  A previous PR attempted to allow single character field names, but there were 2 places validations happened.

## After this PR

Allow single character field names and prevent the issue from happening again by having only 1 source of truth for the validation patterns.

### DIFF
--- a/conjure-core/src/main/java/com/palantir/conjure/parser/types/names/FieldName.java
+++ b/conjure-core/src/main/java/com/palantir/conjure/parser/types/names/FieldName.java
@@ -22,7 +22,6 @@ import com.google.common.base.Preconditions;
 import com.palantir.conjure.CaseConverter;
 import com.palantir.conjure.defs.ConjureImmutablesStyle;
 import com.palantir.conjure.parser.types.complex.ObjectTypeDefinition;
-import java.util.Arrays;
 import org.apache.commons.lang3.StringUtils;
 import org.immutables.value.Value;
 import org.slf4j.Logger;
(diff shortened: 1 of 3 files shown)


conjure-core/src/main/java/com/palantir/conjure/parser/types/names/FieldName.java
```diff
@@ -104,8 +73,8 @@ public static FieldName of(String name) {
     }
 
     /** Converts this {@link FieldName} to a {@link FieldName} with the given case. */
-    public final FieldName toCase(Case targetCase) {
-        return FieldName.of(nameCase().convertTo(this, targetCase));
+    public final FieldName toCase(CaseConverter.Case targetCase) {
+        return FieldName.of(nameCase().convertTo(this.name(), targetCase));
```
aborochoff: Need to validate this.name() is actually what we want here.

aborochoff:
@dansanduleac @iamdanfox fyi

conjure-core/src/main/java/com/palantir/conjure/parser/types/names/FieldName.java
dansanduleac: It is - https://github.com/palantir/conjure/pull/234/files#diff-54cb02b432715a4757215e1539a57a71L67

conjure-core/src/main/java/com/palantir/conjure/parser/types/names/FieldName.java
```diff
@@ -38,60 +37,30 @@
...
         Preconditions.checkArgument(
-                Case.LOWER_CAMEL_CASE.pattern.matcher(name()).matches()
-                        || Case.KEBAB_CASE.pattern.matcher(name()).matches()
-                        || Case.SNAKE_CASE.pattern.matcher(name()).matches(),
+                CaseConverter.Case.LOWER_CAMEL_CASE.getPattern().matcher(name()).matches()
+                        || CaseConverter.Case.KEBAB_CASE.getPattern().matcher(name()).matches()
+                        || CaseConverter.Case.SNAKE_CASE.getPattern().matcher(name()).matches(),
                 "FieldName \"%s\" must follow one of the following patterns: %s",
-                name(), Arrays.toString(Case.values()));
+                name(), Arrays.toString(CaseConverter.Case.values()));
```
dansanduleac: This is a bit dodgy that we're listing them explicily up there, but grabbing the list from CaseConverter here.
Can we have one or the other? Either re-define the list as a private static here, or always use `CaseConverter.Case.values()`

PR #244, opened 2019-03-04
[fix] add docs on endpoint deprecation

ferozco:
## Before this PR
We did not have any documentation on the `deprecated` field on endpoint definitions

## After this PR
Now we do!


cakofony:
👍

PR #258, opened 2019-03-18
RFC: IR format versioning

iamdanfox:
**[Rendered](https://github.com/palantir/conjure/blob/dfox/ir-format-versioning/docs/rfc/007-ir-format-versioning.md)**

## Before this PR

We're somewhat blocked on all the issues tagged ['IR change'](https://github.com/palantir/conjure/labels/IR%20change) because the mechanism for actually making an IR change is so-far undefined.

## After this PR
==COMMIT_MSG==
RFC: IR format versioning recommends single-digit scheme
==COMMIT_MSG==

## Possible downsides?
A bad decision about IR versioning could result in the conjure ecosystem becoming confusing to use / contribute to, or alternatively introduce excessive overhead and unintentionally hamper development.

The upside is that this RFC can always be superseded if significant problems arise.


docs/rfc/007-ir-format-versioning.md
```diff
@@ -0,0 +1,62 @@
...
+
+A contradiction arises because a major.minor versioning scheme can only really satisfy IR consumers or IR producers, it can't meaningfully communicate breaks to both:
+
+
+&nbsp;                     | Additive IR Change     | Restrictive IR Change |
+-------------------------- | ---------------------- | --------------------- |
+**Producers**              | Don't care (minor rev) | Breaking (major rev)
+**Consumers**              | Breaking (major rev)   | Don't care (minor rev)
+
+Producers don't care about additive changes to the IR because they can keep taking upgrades whether the new features are relevant to them or not.  Producers do care about restrictive changes to the IR as this could potentially make their current API definitions invalid, thereby even . This suggests additive changes should increment minor, and restrictive should increment major.
```
cakofony: > thereby even .

I think we're missing some words

cakofony:
This looks sane to me. It reinforces the decision to support a single integer version field, and I don't see serious downsides. If we were to fundamentally change the meaning of the IR we could also rename the version, causing all existing non-compliant parsers/generators to fail as we would expect in alternative version proposals.

PR #276, opened 2019-04-09
[improvement] EndpointDefinition validation errors include the endpoint name and path

carterkozak:
## Before this PR
Previously it was difficult to tell which endpoint caused validation
to fail.

## After this PR
==COMMIT_MSG==
EndpointDefinition validation errors include the endpoint name and path
==COMMIT_MSG==


ferozco:
👍

PR #292, opened 2019-04-25
[Fix] Clarify what parameter types are allowed for path/query/header

uschi2000:
## Before this PR
The spec does not reflect the intent and the reality as per https://github.com/palantir/conjure/blob/master/conjure-core/src/main/java/com/palantir/conjure/defs/validator/EndpointDefinitionValidator.java .

## After this PR
The spec matches the intent and the reality as per https://github.com/palantir/conjure/blob/master/conjure-core/src/main/java/com/palantir/conjure/defs/validator/EndpointDefinitionValidator.java .

### DIFF
--- a/docs/spec/conjure_definitions.md
+++ b/docs/spec/conjure_definitions.md
@@ -357,10 +357,9 @@ An object representing an argument to an endpoint.
 Field | Type | Description
 ---|:---:|---
 type | [ConjureType][] | **REQUIRED**. The type of the value of the argument. The type name MUST exist within the Conjure definition. If this ArgumentDefinition has a param-type of `body` then there are no restrictions on the type. 
-If the param-type is `path` then the de-aliased type MUST be an enum or a primitive (except binary).
-If the param-type is `query` then the de-aliased type MUST be an enum or a primitive (except binary), or a container (list, set, optional) of one of these.
+If the param-type is `path` then the de-aliased type MUST be an enum or a primitive (except `binary`, and `bearertoken`).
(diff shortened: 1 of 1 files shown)


uschi2000:
ping, @iamdanfox

docs/spec/conjure_definitions.md
```diff
@@ -356,7 +356,11 @@ An object representing an argument to an endpoint.
 
 Field | Type | Description
 ---|:---:|---
-type | [ConjureType][] | **REQUIRED**. The type of the value of the argument. The type name MUST exist within the Conjure definition. If this ArgumentDefinition has a param-type of `body` then there are no restrictions on the type. If the param-type is `path`, `header` or `query` then the de-aliased type MUST be an enum, a primitive or an optional of one of these. `query` type parameters may accept lists and sets in addition to optional containers. Additionally, the type MUST NOT be `bearertoken` when the param-type is `header` or `query`.
+type | [ConjureType][] | **REQUIRED**. The type of the value of the argument. The type name MUST exist within the Conjure definition. If this ArgumentDefinition has a param-type of `body` then there are no restrictions on the type. 
+If the param-type is `path` then the de-aliased type MUST be an enum or a primitive.
```
carterkozak: non-binary primitive

uschi2000: done

carterkozak:
:+1:

ferozco:
Thanks for the clarification!

PR #296, opened 2019-05-02
Updated docs to remove deprecated from argument

maclockard:
Appears as though this is no longer a part of [`ArugmentDefinition`](https://github.com/palantir/conjure/blob/master/conjure-core/src/main/java/com/palantir/conjure/parser/services/ArgumentDefinition.java)


carterkozak:
Nice catch

carterkozak:
👍

PR #299, opened 2019-05-04
[improvement] Allow legacy kebab-case and snake_case query params

### LINKED ISSUE #48: Consistent name enforcement for param id
Currently we [allow](https://github.com/palantir/conjure/blob/5147e9bfbe134cc5c13a3a60f8567ab7a73efbfb/conjure-core/src/main/java/com/palantir/conjure/defs/validator/FieldNameValidator.java#L55-L58) fields names to be non-camel case for ease of migration. We however enforce that endpoint parameters [must](https://github.com/palantir/conjure/blob/develop/conjure-core/src/main/java/com/palantir/conjure/defs/validator/EndpointDefinitionValidator.java#L250) be camelCase.

We should be consistent and permissive and allow parameters to also be non-camel case

bmoylan:
Fixes #48

## Before this PR

Validation for query parameters is more strict than for JSON field names.
## After this PR
Endpoint validation allows legacy kebab-case and snake_case query params but logs a warning.

## Possible downsides?
New APIs may accidentally use legacy argument formats. This is mitigated with a logged warning.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/conjure/299)

### DIFF
--- a/conjure-core/src/main/java/com/palantir/conjure/defs/validator/EndpointDefinitionValidator.java
+++ b/conjure-core/src/main/java/com/palantir/conjure/defs/validator/EndpointDefinitionValidator.java
@@ -18,6 +18,7 @@ package com.palantir.conjure.defs.validator;
 
 import com.google.common.base.Preconditions;
 import com.google.common.collect.Sets;
+import com.palantir.conjure.CaseConverter;
 import com.palantir.conjure.either.Either;
 import com.palantir.conjure.spec.ArgumentDefinition;
 import com.palantir.conjure.spec.ArgumentName;
@@ -38,6 +39,7 @@ import com.palantir.conjure.visitor.DealiasingTypeVisitor;
 import com.palantir.conjure.visitor.ParameterTypeVisitor;
 import com.palantir.conjure.visitor.TypeDefinitionVisitor;
 import com.palantir.conjure.visitor.TypeVisitor;
+import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
@@ -69,8 +71,6 @@ public enum EndpointDefinitionValidator implements ConjureContextualValidator<En
         }
     }
 
-    public static final String PATTERN = "[a-z][a-z0-9]*([A-Z0-9][a-z0-9]+)*";
-    public static final Pattern ANCHORED_PATTERN = Pattern.compile("^" + PATTERN + "$");
     public static final Pattern HEADER_PATTERN = Pattern.compile("^[A-Z][a-zA-Z0-9]*(-[A-Z][a-zA-Z0-9]*)*$");
 
     private final ConjureContextualValidator<EndpointDefinition> validator;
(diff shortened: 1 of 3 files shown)


conjure-core/src/main/java/com/palantir/conjure/defs/validator/EndpointDefinitionValidator.java
```diff
@@ -367,30 +367,33 @@ public void validate(EndpointDefinition definition) {
...
         public void validate(EndpointDefinition definition) {
             definition.getArgs().forEach(arg -> {
-                final Pattern pattern;
                 ParameterType paramType = arg.getParamType();
-                if (paramType.accept(ParameterTypeVisitor.IS_BODY)
-                        || paramType.accept(ParameterTypeVisitor.IS_PATH)
-                        || paramType.accept(ParameterTypeVisitor.IS_QUERY)) {
-                    pattern = ANCHORED_PATTERN;
+                if (paramType.accept(ParameterTypeVisitor.IS_BODY) || paramType.accept(ParameterTypeVisitor.IS_PATH)) {
+                    // No validation
```
ferozco: We should still validate body and path params

bmoylan: Previously we would set the `pattern` variable for path/body params but never use it (the if statements below only matched header and query). I assume this is because path and body params don’t have a serialized key, so the param ID is irrelevant

bmoylan: I can make this clearer in the comment

ferozco:
👍

bmoylan:
@ferozco ever seen policy bot fail with this error? I'm wondering if it's because I wasn't logged into external policy-bot before
> failed to compute approval status: failed to list commits: failed to load commit pushed dates: non-200 OK status code: 401 Unauthorized body: "{\"message\":\"Bad credentials\",\"documentation_url\":\"https://developer.github.com/v4\"}"

bmoylan:
Welp, I guess the comment retriggered it and worked

bmoylan:
@ferozco could we tag a release?

conjure-core/src/main/java/com/palantir/conjure/defs/validator/EndpointDefinitionValidator.java
```diff
@@ -367,30 +367,33 @@ public void validate(EndpointDefinition definition) {
...
-                            arg.getParamType(), pattern, paramId.get(), describe(definition));
+                    Preconditions.checkState(
+                            CaseConverter.CAMEL_CASE_PATTERN.matcher(paramId.get()).matches()
+                                    || CaseConverter.KEBAB_CASE_PATTERN.matcher(paramId.get()).matches()
+                                    || CaseConverter.SNAKE_CASE_PATTERN.matcher(paramId.get()).matches(),
+                            "Query param id %s on endpoint %s must match one of the following patterns: %s",
+                                    paramId.get(), describe(definition), Arrays.toString(CaseConverter.Case.values()));
+
+                    if (!CaseConverter.CAMEL_CASE_PATTERN.matcher(paramId.get()).matches()) {
+                        log.warn("Query param ids should be camelCase. kebab-case and snake_case are supported for "
```
uschi2000: what's the plan for moving towards the declared goal state here?

PR #317, opened 2019-06-03
Delete redundant metrics

iamdanfox:
## Before this PR

We had some leftover code that recorded information about conjure definitions, which was useful when we had a test that ran on _all_ our internal definitions, but this test was deleted when we open-sourced.

## After this PR
==COMMIT_MSG==
The conjure CLI no longer depends on dropwizard-metrics
==COMMIT_MSG==


carterkozak:
👍

Issue #316, opened 2019-06-01
Compiled IR should be 100% deterministic


bmoylan:
In godel-conjure-plugin, we're seeing ordering of type definitions in the IR change across runs. Most, but not all, of the time, it matches the order in the yaml definition. Unfortunately I don't have a reliable repo.

This is causing generated code to be unstable (given that multiple types are generated into a single `structs.conjure.go`) and fail our file checksum-based validation.

https://github.com/palantir/godel-conjure-plugin/issues/26

iamdanfox:
When designing the IR, we actually tried to use JSON lists instead of JSON maps wherever possible to avoid this exact problem - if it's just the `types` section of the IR that can get re-ordered though, I think we could put something in that sorts these before emitting them!

markelliot:
Conjure maps should be deterministically ordered if they're run through Conjure objects (at least for Java) because we use LinkedHashMaps instead of plain HashMaps, did we lose that guarantee along the way?

carterkozak:
I think our conjure types with maps retain order if they're built using java builder APIs, however the `JsonSetter` annotated builder methods take `Map` which is a `HashMap` by default, so order has already been lost.

markelliot:
Jackson's default Map type is also LinkedHashMap

iamdanfox:
@markelliot I don't think we've lost the LinkedHashMaps thing you mentioned - I think this specific ticket is saying the YML -> IR process should always produce byte-for-byte identical IR.

carterkozak:
I stand corrected!

markelliot:
@iamdanfox yeah -- I guess the thing that's concerning is that things don't round-trip, and we would expect them to for the Java implementation.

iamdanfox:
I don't think this ticket is alleging that things don't round trip - I think @bmoylan was just saying that for consistent YML, the conjure compiler should always produce the same IR.

markelliot:
That would be true if we used LinkedHashMaps everywhere, and so my concern is that this is signal that we're not doing that.

PR #318, opened 2019-06-03
[fix] Emit IR with types, services and errors sorted by type name

iamdanfox:
## Before this PR

As mentioned in https://github.com/palantir/conjure/issues/316, within the List<TypeDefinition>, the order was actually undefined.

## After this PR
==COMMIT_MSG==
The conjure CLI will always emit IR with types, services and errors sorted alphabetically by their type name.

We don't sort endpoints or arguments within these definitions, instead prefer to preserve the original source ordering.
==COMMIT_MSG==

fixes https://github.com/palantir/conjure/issues/316

## Possible downsides?

cc @bmoylan

### DIFF
--- a/conjure-core/src/main/java/com/palantir/conjure/parser/NormalizeDefinition.java
+++ b/conjure-core/src/main/java/com/palantir/conjure/parser/NormalizeDefinition.java
@@ -45,8 +45,8 @@ public final class NormalizeDefinition {
     }
 
     private static List<ServiceDefinition> sortServiceDefinitions(List<ServiceDefinition> services) {
-        // we intentionally don't sort the Endpoints _within_ a ServiceDefinition, because these can be preserved
-        // from the source yml
+        // we intentionally don't sort the Endpoints _within_ a ServiceDefinition, because we want to preserve
+        // the endpoint order that the API author wrote in their source.yml
         return services.stream()
                 .sorted(Comparator.comparing(def -> typeNameToString(def.getServiceName())))
                 .collect(Collectors.toList());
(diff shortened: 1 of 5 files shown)


conjure-core/src/main/java/com/palantir/conjure/parser/NormalizeDefinition.java
```diff
@@ -0,0 +1,66 @@
...
+
+    private static List<TypeDefinition> sortTypeDefinitions(List<TypeDefinition> types) {
+        return types.stream()
+                .sorted(Comparator.comparing(def -> typeNameToString(def.accept(TypeDefinitionVisitor.TYPE_NAME))))
+                .collect(Collectors.toList());
+    }
+
+    private static List<ServiceDefinition> sortServiceDefinitions(List<ServiceDefinition> services) {
+        // we intentionally don't sort the Endpoints _within_ a ServiceDefinition, because these can be preserved
+        // from the source yml
```
dansanduleac: can we say "we want to preserve the order in the source yml"? the current phrasing doesn't make it clear that we have an opinion on this

conjure-core/src/main/java/com/palantir/conjure/parser/NormalizeDefinition.java
```diff
@@ -0,0 +1,66 @@
...
+
+    private static List<ErrorDefinition> sortErrorDefinitions(List<ErrorDefinition> errors) {
+        return errors.stream()
+                .sorted(Comparator.comparing(def -> typeNameToString(def.getErrorName())))
+                .collect(Collectors.toList());
+    }
+
+    private static String typeNameToString(TypeName typeName) {
+        return typeName.getPackage() + "." + typeName.getName();
+    }
```
dansanduleac: generally prefer something like `comparing(::getPackage).thenComparing(::getName)` rather than making a new string each time

bmoylan:
thanks!

Issue #111, opened 2018-10-01
[Discussion] Remove `type` field from union types


raiju:
Conjure's union types can end up being extremely inefficient in terms of bytes in the wire format, for example when encoding ASTs (where there is generally a lot of polymorphic branching). The double-encoding of the type for union types is the main first order cause, as for small-ish ASTs it will almost double the size:
```
Lit(Int(42)):
{"type":"lit","lit":{"type":"int","int":42}}
vs
{"lit":{"int":42}}
```
while not actually adding any new information. @markelliot believes the second form can be implemented for all currently-known targets.

Has this already been discussed? Is this worth the extra work to change?

@markelliot We discussed this a little bit ago, in the context of the Conjure encoding overhead (in terms of bytes used). As it will be significantly more expensive to make this change once this specification has been released (as it's a wire format change), I wanted to start a discussion now.

ericanderson:
Why not do something more like `{ __type: string, ...typesProperties }`?

sfackler:
That either requires us to guarantee that the __type field comes first in the serialized form, which can be hard to enforce through various layers of glue, or buffer all of the properties until we figure out what type to deserialize into.

cakofony:
That would not work for primitive types, not everything has properties.

markelliot:
All unions encode the value in a specific field name, and the field name can be used to identify the type of its value (defitionally). 

One might model union types as a shorthand for a single type with many optional fields and a guarantee that exactly one of those fields is non-empty (and thus present in the wire format).

tom-s-powell:
Has there been any further discussion on this topic? Have a use-case where users will be constructing JSON queries that heavily involve union types and the non-type proposal would be far less verbose. Could something like this be a feature flag that a service can configure based on what best fits the use-case and intended client?

markelliot:
@tom-s-powell I think we'd need to do some work to ensure that all deserializers across all the languages get to the point where `type` is ignored and then do a major rev on the protocol in order to stop producing the extra key. To my knowledge, none of the languages have the requisite changes in to use the field key as a differentiator for type instead of examining the type key..

sfackler:
Rust implementation: https://github.com/palantir/conjure-rust/commit/181e0a665c0d0c20de7866889dda7d873eb25f60

PR #422, opened 2019-10-22
Fix: Don't allow a union variant called 'type'

### LINKED ISSUE #179: `type` should be forbidden as a union variant name
This definition passes validation and generates code with the conjure-java backend, but can't possibly work:

```yaml
types:
  definitions:
    objects:
      union:
        type: string
```

It should be forbidden by the conjure parser.

dansanduleac:
## Before this PR

`type` was an acceptable union variant name, but we have no idea how to serialize it to JSON. We can't use `"type"` as that's already reserved for the chosen union variant name.

This edge case is not mentioned in [the yml definition spec](https://palantir.github.io/conjure/#/docs/spec/conjure_definitions?id=uniontypedefinition) nor in [the wire format](https://palantir.github.io/conjure/#/docs/spec/wire?id=_54-union-json-format).

## After this PR
==COMMIT_MSG==
Don't allow a union variant called `type` as we don't know how that should be serialized.
==COMMIT_MSG==

Fixes #179.

## Possible downsides?

People might have come up with a format for how to serialize this edge case in a particular generator only (e.g. go, rust?) that didn't make it back to the spec, in which case we'd break those use cases.

### DIFF
--- a/changelog/@unreleased/pr-422.v2.yml
+++ b/changelog/@unreleased/pr-422.v2.yml
@@ -1,0 +1,6 @@
+type: fix
+fix:
+  description: Don't allow a union variant called `type` as we don't know how that
+    should be serialized.
+  links:
+  - https://github.com/palantir/conjure/pull/422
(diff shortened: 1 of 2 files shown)


sfackler:
If #111 ever lands we could un-forbid this.

PR #423, opened 2019-10-25
Fix #325: Validate map keys have PLAIN representation

### LINKED ISSUE #325: Map key must be builtin non-any type
## What happened?

In the wire spec, we imply that [map keys may only be types that have a `PLAIN` representation](https://palantir.github.io/conjure/#/docs/spec/wire?id=_52-container-types):
> A key k is serialized as a string with contents PLAIN(k).

And indeed this is what our generators expect, because we represent conjure maps as JSON objects, and JSON object keys must be strings.

We don't mention this requirement in the [Conjure definitions section](https://palantir.github.io/conjure/#/docs/spec/conjure_definitions?id=containertype), and the conjure compiler does not enforce it.

We should mention this constraint in the Conjure definitions section, and fail to compile code that violates it.

carterkozak:
==COMMIT_MSG==
Validate map keys have PLAIN representation

Previously it was possible to build a conjure definition for a
map that was impossible to serialize as JSON. If this fails
on an external type import, please make sure you have
selected the most accurate fallback type value, the default
`any` is not allowed.
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-423.v2.yml
+++ b/changelog/@unreleased/pr-423.v2.yml
@@ -1,0 +1,12 @@
+type: improvement
+improvement:
+  description: |-
+    Validate map keys have PLAIN representation
+
+    Previously it was possible to build a conjure definition for a
+    map that was impossible to serialize as JSON. If this fails
+    on an external type import, please make sure you have
+    selected the most accurate fallback type value, the default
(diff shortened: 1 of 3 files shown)


dansanduleac:
👍

PR #426, opened 2019-10-28
Revert "Add 401 Unauthenticated to conjure error types"

dansanduleac:
Reverts palantir/conjure#367

Unforetunately I merged this too early.

The consensus of the Infra team is that we want this, but we need to figure out #258 first, otherwise we'll never be able to cut a release.

Currently, this being on develop blocks us from releasing the other hotifxes on develop.


ferozco:
👍

PR #439, opened 2019-11-06
Fix NamedTypesDefinition typo in Conjure spec

tdeitch:
...at least I think it's a typo, please sanity check me :)


ferozco:
doze?

PR #494, opened 2020-02-03
Fix #120: Enum values, fields, and union values support deprecation

### LINKED ISSUE #120: Support for field and enum-value markers/deprecations in conjure definition
It would be useful to add the ability to specify markers on individual fields and enum values to conjure definitions. 

The initial use case that this request is deprecating fields and enum values. In models or enums that are used widely, it may not be feasible to create full replacements (e.g. Person2 replaces Person). By marking a field deprecated, tooling could be used to ensure deprecated fields are removed following major releases, as an example.

Workarounds such as adding keywords in docs would work, but may clutter up useful documentation and wouldn't be portable.

carterkozak:
## Before this PR
No deprecation docs on fields, union types, or enum values.

## After this PR
==COMMIT_MSG==
 Enum values, fields, and union values support deprecation documentation.
==COMMIT_MSG==

## Possible downsides?

This is an additive change to the intermediate representation,
however there is little risk for consumers because this information
is simple metadata.

### DIFF
--- a/changelog/@unreleased/pr-494.v2.yml
+++ b/changelog/@unreleased/pr-494.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Enum values, fields, and union values support deprecation documentation.
+  links:
+  - https://github.com/palantir/conjure/pull/494
--- a/conjure-api/src/main/conjure/conjure-api.yml
+++ b/conjure-api/src/main/conjure/conjure-api.yml
@@ -129,11 +129,13 @@ types:
         fields:
           value: string
           docs: optional<Documentation>
+          deprecated: optional<Documentation>
       FieldDefinition:
         fields:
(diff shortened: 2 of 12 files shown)


ferozco:
After discussing offline we decided that making this change does not require a major rev of the IR since it is just an additive change that does not impact the the semantics of the API.

I also validated that the infra owned generators (Java, Python, TypeScript) do not fail on unknown fields making the change safe from a backwards compatability perspective for those generators. @sfackler @nmiyake could you guys please verify that this change would not cause issues with your generators?

sfackler:
conjure-rust currently rejects unknown fields, but I just opened a PR to change that: https://github.com/palantir/conjure-rust/pull/98

nmiyake:
Verified that `conjure-go` handles this extra field without issue.

PR #497, opened 2020-02-04
Fix #491: Validate against binary data in non-body container parameters

### LINKED ISSUE #491: Conjure generator accepts invalid query parameter types
## What happened?
Ran the Conjure generator on the following definition, and it succeeded/generated IR without any errors:

```yml
services:
  TestService:
    name: Test Service
    package: api
    endpoints:
      queryParamListBinary:
        http: GET /binaryListQueryVar
        args:
          myQueryParam1:
            type: list<binary>
            param-type: query
```

CLI should report that the definition is invalid/illegal and fail, as, per the [Conjure definition document](https://github.com/palantir/conjure/blob/master/docs/spec/conjure_definitions.md#argumentdefinition), query parameters cannot be binary (or a container of it):

> If the param-type is query then the de-aliased type MUST be an enum or a primitive (except binary and bearertoken), or a container (list, set, optional) of one of these.

carterkozak:
## Before this PR
Query and header parameters using `list<binary>` `set<binary>` and `optional<binary>` types incorrectly passed validation.

## After this PR
==COMMIT_MSG==
Validate against binary data in non-body container parameters.
==COMMIT_MSG==

### DIFF
--- a/conjure-core/src/main/java/com/palantir/conjure/defs/validator/EndpointDefinitionValidator.java
+++ b/conjure-core/src/main/java/com/palantir/conjure/defs/validator/EndpointDefinitionValidator.java
@@ -100,12 +100,10 @@ public enum EndpointDefinitionValidator implements ConjureContextualValidator<En
             definition.getArgs()
                     .stream()
                     .filter(arg -> !arg.getParamType().accept(ParameterTypeVisitor.IS_BODY))
-                    .forEach(arg -> {
-                        Preconditions.checkArgument(
-                                validateType(arg.getType(), dealiasingTypeVisitor),
-                                "Non body parameters cannot be of the 'binary' type: '%s' is not allowed",
-                                arg.getArgName());
-                    });
+                    .forEach(arg -> Preconditions.checkArgument(
+                            validateType(arg.getType(), dealiasingTypeVisitor),
+                            "Non body parameters cannot be of the 'binary' type: '%s' is not allowed",
+                            arg.getArgName()));
         }
 
         private static boolean validateType(Type input, DealiasingTypeVisitor dealiasingTypeVisitor) {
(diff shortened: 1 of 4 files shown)


conjure-core/src/main/java/com/palantir/conjure/defs/validator/EndpointDefinitionValidator.java
```diff
@@ -100,17 +101,32 @@ public void validate(EndpointDefinition definition, DealiasingTypeVisitor dealia
...
+            if (input.accept(TypeVisitor.IS_OPTIONAL)) {
+                return validateType(input.accept(TypeVisitor.OPTIONAL).getItemType(), dealiasingTypeVisitor);
+            }
+            if (input.accept(TypeVisitor.IS_LIST)) {
+                return validateType(input.accept(TypeVisitor.LIST).getItemType(), dealiasingTypeVisitor);
+            }
+            if (input.accept(TypeVisitor.IS_SET)) {
+                return validateType(input.accept(TypeVisitor.SET).getItemType(), dealiasingTypeVisitor);
+            }
+            return !type.accept(TypeVisitor.IS_BINARY) && !type.accept(TypeVisitor.IS_ANY);
```
nmiyake: Not sure if this should be handled in a separate PR, but for query and path parameters, the check should also reject containers that contain `bearertoken`.

conjure-core/src/main/java/com/palantir/conjure/defs/validator/EndpointDefinitionValidator.java
```diff
@@ -100,17 +101,32 @@ public void validate(EndpointDefinition definition, DealiasingTypeVisitor dealia
...
-                        boolean isValid = dealiasingTypeVisitor.dealias(arg.getType())
-                                .fold(
-                                        typeDefinition -> true,
-                                        type -> !type.accept(TypeVisitor.IS_BINARY)
-                                                && !type.accept(TypeVisitor.IS_ANY)
-                                );
                         Preconditions.checkArgument(
-                                isValid, "Non body parameters cannot be of the 'binary' type: '%s' is not allowed",
+                                validateType(arg.getType(), dealiasingTypeVisitor),
+                                "Non body parameters cannot be of the 'binary' type: '%s' is not allowed",
```
nmiyake: "cannot reference the 'binary' type" might be a bit more accurate, since `list<binary>` is technically not the binary type.

nmiyake: Would it also be possible to print the actual type definition in the error message (`list<binary>` etc.)? If so, I think that would make the error message a bit more user-friendly.

carterkozak: Good call. Rendering the human-readable type definition isn't trivial, but I've added the endpoint name and path for easier debugging.

conjure-core/src/main/java/com/palantir/conjure/defs/validator/EndpointDefinitionValidator.java
```diff
@@ -99,17 +100,32 @@ public void validate(EndpointDefinition definition, DealiasingTypeVisitor dealia
...
+
+        private static boolean validateType(Type input, DealiasingTypeVisitor dealiasingTypeVisitor) {
+            Optional<Type> dealiased = dealiasingTypeVisitor.dealias(input)
+                    .fold(typeDefinition -> Optional.empty(), Optional::of);
+            // typeDef isn't binary
+            if (!dealiased.isPresent()) {
+                return true;
+            }
+            Type type = dealiased.get();
+            if (input.accept(TypeVisitor.IS_OPTIONAL)) {
```
iamdanfox: rather than a chain of if statements here, shall we just use a Type.Visitor so that it's guaranteed to be exhaustive?

carterkozak: The visitor creates a lot of duplication with the dealiasing visitor and requires us to handle types that are already disallowed elsewhere (map). I think the current code is easier to follow, but it would be more difficult to find if we added new container types.

iamdanfox: OK fair enough - guess we're not at particular risk from that as no new collection types coming down the pipeline!

carterkozak:
Dozer?

PR #532, opened 2020-02-19
add extensions to conjure

ferozco:
## Before this PR
There was no way of adding metadata to a Conjure definitions.

## After this PR
==COMMIT_MSG==
Add `extensions` field to conjure definition which stores metadata about the definition itself. 

Importantly, the fields stored in metadata should not impact the semantics of the API. Expected usage would be things like origin, version, product dependencies etc.
==COMMIT_MSG==

## Possible downsides?
This is a grab bag that allows anyone to stuff whatever they would like into a conjure extension.


iamdanfox:
I think we need some sentences to describe the intentions of this in the markdown document.

conjure-api/src/main/conjure/conjure-api.yml
```diff
@@ -5,6 +5,7 @@ types:
       ConjureDefinition:
         fields:
           version: integer
+          extensions: map<string, any>
```
iamdanfox: could we re-order this so that it goes types, services, errors, extensions?

iamdanfox:
Could we also establish a convention about the 'keys' of the extensions block? like should it include a generator?

ferozco:
> I think we need some sentences to describe the intentions of this in the markdown IR document.

sure, i'll add something

> Could we also establish a convention about the 'keys' of the extensions block? like should it include a generator?

I think by adding a map<string, any> we're basically saying that outside of having no semantic meaning to the API conjure is opinionated about what goes into the map. I'm also not sure what a good convention would be given that the primary use case (product-dependencies) is generator agnostic as are other ideas (origin repository, version, name)

conjure-api/src/main/conjure/conjure-api.yml
```diff
@@ -8,6 +8,7 @@ types:
           errors: list<ErrorDefinition>
           types: list<TypeDefinition>
           services: list<ServiceDefinition>
+          extensions: map<string, any>
```
iamdanfox: I kinda want to put a floating type in the IR which is a ProductDependency, that people can json convertValue one of these objects into... otherwise we'll be re-defining the same thing so many times

docs/spec/intermediate_representation.md
```diff
@@ -476,6 +476,27 @@ for details on the field definition format.
...
+
+Example extension definition:
+```json
+{
+  "extensions": {
+    "product-dependencies": [
+      {
+        "minimum-version": "1.0.0",
+        "recommended-version": "1.1.0",
+        "maximum-version": "1.x.x"
```
iamdanfox: the dashes are not conjure expressible - could we go camel?

docs/spec/intermediate_representation.md
```diff
@@ -476,6 +476,27 @@ for details on the field definition format.
...
+
+The "extensions" section is a map of named extensions. The structure of each extension is user defined and
+any consumer of an extension should treat its presence as optional and be tolerant of unknown fields. Extensions must
+not modify the semantics of definitions and are intended only for conveying metadata about the definition.
+
+Example extension definition:
+```json
+{
+  "extensions": {
+    "product-dependencies": [
```
iamdanfox: should this be sls-recommended-product-depndencies to match what we put in jar manifests?

conjure-api/src/main/conjure/conjure-api.yml
ferozco: I'd prefer to not add the type to Conjure since in practice you'd want language specific bindings for `SlsVersion` and `SlsVersionMatcher`. It also re-enforces that things within the extension are outside of the spec

docs/spec/intermediate_representation.md
ferozco: I was going for consistency with what is included in manifests (JAR, and SLS)

docs/spec/intermediate_representation.md
ferozco: sure, I'll update https://github.com/palantir/gradle-conjure/pull/381 to use `sls-recommended-product-dependencies`

docs/spec/intermediate_representation.md
iamdanfox: I think i'd prefer conjure to be internally consistent, and accept that this makes us inconsistent with the manifests out there in the wild... at least with jackson we can easily deserialize both but serialize as the dashes so you don't need two classes?

conjure/src/test/java/com/palantir/conjure/cli/ConjureCliTest.java
```diff
@@ -46,10 +47,15 @@ public void before() throws IOException {
 
     @Test
     public void correctlyParseArguments() {
-        String[] args = {"compile", inputFile.getAbsolutePath(), outputFile.getAbsolutePath()};
+        String[] args = {
+                "compile",
+                inputFile.getAbsolutePath(),
+                outputFile.getAbsolutePath(),
+                "--extensions", "{\"foo\": \"bar\"}"};
```
iamdanfox: thoughts on passing the extensions as a file instead? it might make it easier to debug if you could open up a directory locally and look at the json?

ferozco: I actually thought about that but preferred the inline JSON since it makes it more usable as a CLI. It also simplifies the gradle conjure implementation. From a debugability perspective, we log all of the arguments being passed to the compile in gradle and so it should be easy understand what was sent over

docs/spec/intermediate_representation.md
ferozco: I'd prefer to present a different example in the spec than change the structure of the product dependencies in production

iamdanfox: I'm just saying we've worked so hard to try and make conjure this nice clean world where things are all consistent, and I don't want to compromise that over the camel vs kebab thing. Not saying we need to change what's used in production, I was thinking that we can just do a translation?

conjure/src/test/java/com/palantir/conjure/cli/ConjureCliTest.java
iamdanfox: ok fair enough

iamdanfox:
I think this looks ok and unblocks what we need, I'm just having a v. strong reaction to the kebabs.

PR #533, opened 2020-02-19
Conjure CLI is permissive with unknown options

ferozco:
## Before this PR
Conjure CLI would barf if any unknown options were provided. This made adding a new option require a lockstep upgrade between the CLI and the build tooling that would populate the option

## After this PR
==COMMIT_MSG==
Conjure CLI will log and ignore any unknown options that are provided
==COMMIT_MSG==

## Possible downsides?
N/A

### DIFF
--- a/conjure/src/main/java/com/palantir/conjure/cli/ConjureCli.java
+++ b/conjure/src/main/java/com/palantir/conjure/cli/ConjureCli.java
@@ -22,12 +22,9 @@ import com.fasterxml.jackson.datatype.jdk8.Jdk8Module;
 import com.google.common.annotations.VisibleForTesting;
 import com.palantir.conjure.defs.Conjure;
 import com.palantir.conjure.spec.ConjureDefinition;
-import com.palantir.logsafe.SafeArg;
 import java.io.IOException;
 import java.util.List;
 import javax.annotation.Nullable;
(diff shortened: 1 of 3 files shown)


ferozco:
In support of #532

conjure/src/main/java/com/palantir/conjure/cli/ConjureCli.java
```diff
@@ -60,8 +67,15 @@ public void run() {
...
         private String output;
 
+        @CommandLine.Unmatched
+        @Nullable
+        private List<String> unmatchedOptions;
+
         @Override
         public void run() {
+            if (unmatchedOptions != null && !unmatchedOptions.isEmpty()) {
+                log.warn("Ignoring unknown options", SafeArg.of("unknown options", unmatchedOptions));
```
carterkozak: Do we actually want `System.err.println` here? Using a logger for CLI user interaction can be problematic, for example slf4j-simple won't print the args because we don't have substitution braces.
If slf4j-simple is replaced with something else, it won't be obvious that this doesn't provide user feedback.

ferozco: haha I had originally gone with that then switched over to logging. I'll revert

PR #537, opened 2020-02-21
Better error message telling you which file the error came from

CRogers:
## Before this PR
If your conjure definition file failed validation, the old error message would not tell you what file the error was in:

```
java.lang.IllegalStateException: Endpoint 'testEndpoint{http: GET /path}' cannot have multiple body parameters: [arg, arg2]
```

## After this PR
==COMMIT_MSG==
Errors encountered when parsing or validating conjure definition files will now tell you which file the error was found in.
==COMMIT_MSG==

The error now looks like:

```
java.lang.RuntimeException: Encountered error trying to parse file 'test'
// stacktrace
Caused by: java.lang.IllegalStateException: Endpoint 'testEndpoint{http: GET /path}' cannot have multiple body parameters: [arg, arg2]
```

## Possible downsides?
More complicated code.

### DIFF
--- a/changelog/@unreleased/pr-537.v2.yml
+++ b/changelog/@unreleased/pr-537.v2.yml
@@ -1,0 +1,6 @@
+type: fix
+fix:
+  description: Errors encountered when parsing or validating conjure definition files
+    will now tell you which file the error was found in.
+  links:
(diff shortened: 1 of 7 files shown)


topherinternational:
Awesome, thanks @CRogers!

PR #580, opened 2020-03-27
External import supports binary base type

### LINKED ISSUE #579: Can't use binary as a base type
## What happened?

I tried:

```
types:
  imports:
    Response:
      base-type: binary
      external:
        java: javax.ws.rs.core.Response
```

It failed with:

```
  Caused by: java.lang.IllegalArgumentException: TypeNames must be a primitive type [any, string, integer, double, boolean, safelong, rid, bearertoken, uuid] or match pattern ^[A-Z][a-z0-9]+([A-Z][a-z0-9]+)*$: binary
        at com.google.common.base.Preconditions.checkArgument(Preconditions.java:455)
        at com.palantir.conjure.parser.types.names.TypeName.check(TypeName.java:46)
        at com.palantir.conjure.parser.types.names.ImmutableTypeName.validate(ImmutableTypeName.java:110)
        at com.palantir.conjure.parser.types.names.ImmutableTypeName.access$200(ImmutableTypeName.java:30)
        at com.palantir.conjure.parser.types.names.ImmutableTypeName$Builder.build(ImmutableTypeName.java:197)
        at com.palantir.conjure.parser.types.names.TypeName.of(TypeName.java:54)
        at com.palantir.conjure.parser.types.primitive.PrimitiveType.fromString(PrimitiveType.java:70)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at com.fasterxml.jackson.databind.introspect.AnnotatedMethod.callOnWith(AnnotatedMethod.java:122)
        at com.fasterxml.jackson.databind.deser.std.FactoryBasedEnumDeserializer.deserialize(FactoryBasedEnumDeserializer.java:138)
        ... 64 more
```

I want a binary base type to be supported.

ferozco:
Closes #579 

## After this PR
==COMMIT_MSG==
External imports can have binary as their base type
==COMMIT_MSG==

### DIFF
--- a/conjure-core/src/main/java/com/palantir/conjure/parser/types/names/TypeName.java
+++ b/conjure-core/src/main/java/com/palantir/conjure/parser/types/names/TypeName.java
@@ -22,6 +22,8 @@ import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableSet;
 import com.palantir.conjure.defs.ConjureImmutablesStyle;
 import com.palantir.conjure.parser.types.NamedTypesDefinition;
+import com.palantir.conjure.spec.PrimitiveType;
+import java.util.Arrays;
 import java.util.regex.Pattern;
 import org.immutables.value.Value;
 
@@ -34,9 +36,9 @@ import org.immutables.value.Value;
 public abstract class TypeName {
 
     private static final Pattern CUSTOM_TYPE_PATTERN = Pattern.compile("^[A-Z][a-z0-9]+([A-Z][a-z0-9]+)*$");
-    static final ImmutableSet<String> PRIMITIVE_TYPES =
-            ImmutableSet.of(
-                    "any", "string", "integer", "double", "boolean", "safelong", "rid", "bearertoken", "uuid");
+    private static final ImmutableSet<String> PRIMITIVE_TYPES = Arrays.stream(PrimitiveType.Value.values())
+            .map(value -> value.name().toLowerCase())
+            .collect(ImmutableSet.toImmutableSet());
 
     @JsonValue
     public abstract String name();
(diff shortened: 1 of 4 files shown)


carterkozak:
This would be easy to misuse. We have a ton of special handling for binary streaming around dealiasing that may or may not work as expected with external type imports.

iamdanfox:
Ok so in order to enable @mglazer's to write:
```yml
types:
  imports:
    Response:
      base-type: binary
      external:
        java: javax.ws.rs.core.Response
```

I think we'd need to do some special-casing in the dialogue generator to just ignore external imports with `base-type: binary` and just code-generate `InputStream` instead... this feels a bit odd because everywhere else we've respected people's external imports.

carterkozak:
It breaks conjure-undertow as well

iamdanfox:
I think this is just what happens when you rely on things that are not inside the conjure ecosystem unfortunately :/. @mglazer is there any feasibility to use just a `binary` response type instead of the jaxrs one?

mglazer:
So the reason why I switched to `javax.ws.rs.Response` in the first place is that I needed to set a `Content-Disposition` header so this particularly endpoint would be treated as an attachment endpoint. I agree this shouldn't have been made into a conjure endpoint in the first place if it's just serving an arbitrary file, but this is where we are.

What I've done for now that seems to be the only thing that works is setting `jerseyBinaryAsResponse` as a conjure option, which seems to work pretty well, but I suspect if we ever stop using the Jersey endpoints, something will break?

PR #603, opened 2020-05-12
remove endpoint marker documentation

lycarter:
## Before this PR
Endpoint markers have been deprecated since the release of OSS conjure (?), and certainly aren't supported in conjure-java (see https://github.com/palantir/conjure-java/pull/883)

## After this PR
==COMMIT_MSG==
Misleading documentation about endpoint markers is removed
==COMMIT_MSG==


lycarter:
cc @carterkozak

PR #667, opened 2020-08-11
cache parsed results across multiple runs

rzpt:
## Before this PR

Parsed results were not cached between runs, so certain common files could be parsed many times in a single run.

## After this PR

==COMMIT_MSG==
Each conjure file will only be parsed once
==COMMIT_MSG==

## Possible downsides?

The code is no longer thread safe due to sharing of a static HashMap across threads, but the cli is single-threaded.

### DIFF
--- a/changelog/@unreleased/pr-667.v2.yml
+++ b/changelog/@unreleased/pr-667.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Each conjure file will only be parsed once
+  urls:
+    - https://github.com/palantir/conjure/pull/667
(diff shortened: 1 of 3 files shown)


ferozco:
@rzpt mind giving a bit more details about how you're using Conjure? Given that it is typically used an executable, I don't believe the change you have made will have any impact across "run"s

rzpt:
Sorry, multiple "runs" is not the right word. I meant the loop here: https://github.com/palantir/conjure/blob/master/conjure-core/src/main/java/com/palantir/conjure/defs/Conjure.java#L38

In a certain internal repo, we have many conjure yml files in a single project (117), and certain files are being parsed many times while invoking the gradle conjure-ir task. Many of these files were being parsed several times, with one of them being parsed 40+ times.

rzpt:
It seems like I can't write the changelog by using changelog-app? Can you do that for me?

ferozco:
Ok, rather than introducing static state could we refactor the code to allow us to re-use the Recursive parser?

rzpt:
Done.

rzpt:
@ferozco Can this be merged? I can't write a changelog with the changelog-app. Could you try that?

ferozco:
👍

carterkozak:
I've fixed up the changelog so we can cut a release with this fix. #669

PR #674, opened 2020-08-19
Revert "cache parsed results across multiple runs (#667)"

CRogers:
This reverts commit 5715c5ce4cf0dc7dc5c51de8709c57666f4943ec / #667.

## Before this PR
After 4.11.0, conjure seemingly does nothing at all. If you delete a chunk of your conjure yml file, such that it isn't even valid yaml, then rerun `./gradlew compileConjure`, it does produce an error. If you make a valid change, it does not produce updated conjure IR.

## After this PR
==COMMIT_MSG==
Reverts change to add parsing caching for conjure files, which broke conjure IR outputting.
==COMMIT_MSG==

## Possible downsides?
None, I think, as it does not seem to work. It's kind of worrying we have no test failures (going to check they are actually running!). The caching feature sounds useful if it speeds up the build, so it may be worth giving it another go with tests to ensure it still works.

cc @rzpt.

### DIFF
--- a/changelog/@unreleased/pr-674.v2.yml
+++ b/changelog/@unreleased/pr-674.v2.yml
@@ -1,0 +1,6 @@
+type: fix
+fix:
+  description: Reverts change to add parsing caching for conjure files, which broke
+    conjure IR outputting.
+  links:
+  - https://github.com/palantir/conjure/pull/674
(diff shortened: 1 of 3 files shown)


ferozco:
Do we understand why this problem is happening? It seems pretty odd that caching within the Conjure process would impact output across invocations in gradle.

CRogers:
Hmm, I guess we must not take into account the conjure IR version for caching, for this to have rolled out the way it has.

CRogers:
After diving in with @ferozco, looks like it's actually some problem in gradle-conjure.

PR #682, opened 2020-08-24
fix publish

ferozco:
## After this PR

==COMMIT_MSG==
fix publish
==COMMIT_MSG==


ferozco:
root project was inadvertently renamed in https://github.com/palantir/conjure/pull/486

robert3005:
👍

PR #702, opened 2020-09-17
feature: add per endpoint tags

### LINKED ISSUE #612: non-Java implementations of argument markers
Markers are currently impossible to implement in other languages. The [documentation](https://github.com/palantir/conjure/blob/master/docs/spec/conjure_definitions.md#argumentdefinition) says:

> List of types that serve as additional metadata for the argument. If the value of the field is a string it MUST be a type name that exists within the Conjure definition.

In conjure-java, `markers` on endpoint parameters are interpreted as annotations on method arguments. The most common use case is to mark arguments safe for logging.

I'd argue the spec is overfit to the java implementation. Even if we did have some form of annotation, these are always external types since conjure does not allow defining new annotations. Conjure only supports a `java` key for external types, so there's not really a path forward in other languages.

In conjure-go we'd like to be able to mark arguments safe, but go does not have a similar annotation construct for which there exist discrete types that can be imported into a definition. For example, witchcraft-go-server's endpoint registration uses option parameters like [`wrouter.SafePathParams`](https://pkg.go.dev/github.com/palantir/witchcraft-go-server/wrouter?tab=doc#SafePathParams) to indicate parameter safety.

Are there other markers in use anywhere? Should safety become a first-class concept so language implementations can decide how to deal with them independently?

ferozco:
Closes #612
## Before this PR
users were able to annotate endpoints with type `markers` which was used by Java generator to apply annotations to endpoints or to indicate to the generator to different endpoint signatures (ex. Java-undertow and `Async` type). This was problematic since the type based approach limited the usefulness of markers to Java projects as well as prevented users from adding metadata to their definitions that could be used non-generator consumers of the IR (i.e. Documentation services)

## After this PR

==COMMIT_MSG==
Add `tags` field to `EndpointDefinition` in the IR
==COMMIT_MSG==

Any values provided in this field must not impact the wire structure of the endpoint and should not be considered when evaluating backwards compatability.

## Possible downsides?
Duplicated functionality with `markers` could be confusing and the untyped set of tags leaves a lot of room for potential misuse

### DIFF
--- a/conjure-api/src/main/conjure/conjure-api.yml
+++ b/conjure-api/src/main/conjure/conjure-api.yml
@@ -186,6 +186,7 @@ types:
           paramType: ParameterType
           docs: optional<Documentation>
           markers: list<Type>
+          tags: list<string>
       ArgumentName:
         alias: string
         docs: >
--- a/conjure-api/src/main/conjure/conjure-api.yml
+++ b/conjure-api/src/main/conjure/conjure-api.yml
@@ -158,6 +158,7 @@ types:
           docs: optional<Documentation>
           deprecated: optional<Documentation>
           markers: list<Type>
+          tags: set<string>
       EndpointName:
         alias: string
         docs: Should be in lowerCamelCase.
--- a/conjure-core/src/main/java/com/palantir/conjure/defs/ConjureParserUtils.java
+++ b/conjure-core/src/main/java/com/palantir/conjure/defs/ConjureParserUtils.java
@@ -72,6 +72,7 @@ import com.palantir.conjure.spec.TypeName;
 import com.palantir.conjure.spec.UnionDefinition;
 import com.palantir.conjure.visitor.DealiasingTypeVisitor;
 import com.palantir.conjure.visitor.TypeDefinitionVisitor;
+import com.palantir.logsafe.Preconditions;
 import com.palantir.logsafe.exceptions.SafeIllegalArgumentException;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -363,7 +364,9 @@ public final class ConjureParserUtils {
                 .httpPath(httpPath)
                 .auth(def.auth().map(ConjureParserUtils::parseAuthType).orElse(defaultAuth))
                 .args(parseArgs(def.args(), httpPath, typeResolver))
-                .tags(def.tags())
+                .tags(def.tags().stream()
+                        .peek(tag -> Preconditions.checkArgument(!tag.isEmpty(), "tag must not be empty"))
+                        .collect(Collectors.toSet()))
                 .markers(parseMarkers(def.markers(), typeResolver))
                 .returns(def.returns().map(t -> t.visit(new ConjureTypeParserVisitor(typeResolver))))
                 .docs(def.docs().map(Documentation::of))
@@ -414,7 +417,9 @@ public final class ConjureParserUtils {
                     .paramType(paramType)
                     .docs(original.docs().map(Documentation::of))
                     .markers(parseMarkers(original.markers(), typeResolver))
-                    .tags(original.tags());
+                    .tags(original.tags().stream()
+                            .peek(tag -> Preconditions.checkArgument(!tag.isEmpty(), "tag must not be empty"))
+                            .collect(Collectors.toSet()));
             resultBuilder.add(builder.build());
         }
         return resultBuilder.build();
--- a/conjure-core/src/main/java/com/palantir/conjure/defs/ConjureParserUtils.java
+++ b/conjure-core/src/main/java/com/palantir/conjure/defs/ConjureParserUtils.java
@@ -413,7 +413,8 @@ public final class ConjureParserUtils {
                     .type(original.type().visit(new ConjureTypeParserVisitor(typeResolver)))
                     .paramType(paramType)
                     .docs(original.docs().map(Documentation::of))
-                    .markers(parseMarkers(original.markers(), typeResolver));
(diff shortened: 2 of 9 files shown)


ferozco:
cc @bmoylan @sfackler @nmiyake as generator owners

conjure-core/src/main/java/com/palantir/conjure/parser/services/EndpointDefinition.java
```diff
@@ -35,6 +35,8 @@
 
     Map<ParameterName, ArgumentDefinition> args();
 
+    Set<String> tags();
```
carterkozak, line 38: must be passed to the IR

carterkozak (commented):
Design makes sense. Needs updates to the docs, and the value from conjure yml must make its way to the IR.

carterkozak:
Worth documenting for posterity why we're using `set<string>` instead of `map<string,string>`

conjure-api/src/main/conjure/conjure-api.yml
```diff
@@ -158,6 +158,7 @@ types:
           docs: optional<Documentation>
           deprecated: optional<Documentation>
           markers: list<Type>
+          tags: set<string>
```
carterkozak, line 161: `docs` uses the `Documentation` `alias<string>` type, should we use a string alias here as well?

bmoylan: Is there a reason we don't use an enum if we're defining a new field? It would be nice for this to be more strongly typed, unless the assertion is that different generators might respect specific/different/new/custom tags.

docs/spec/conjure_definitions.md
```diff
@@ -348,6 +348,7 @@ returns | [ConjureType][] | The name of the return type of the endpoint. The val
 args | Map[`string` &rarr; [ArgumentDefinition][]&nbsp;or&nbsp;[ConjureType][]] | A map between argument names and argument definitions. If the value of the field is a `string` it MUST be a type name that exists within the Conjure definition. Furthermore, if a `string` the argument will default to `auto` [ArgumentDefinition.ParamType][].
 docs | [DocString][] | Documentation for the endpoint. [CommonMark syntax](http://spec.commonmark.org/) MAY be used for rich text representation.
 deprecated | [DocString][] | Documentation for the deprecation of the endpoint. [CommonMark syntax](http://spec.commonmark.org/) MAY be used for rich text representation.
+tags | Set[`string`] | Set of tags that serves as additional metadata for the endpoint.
```
bmoylan, line 351: Can we provide any guidance for examples or expected usage? If I saw this without knowing the context, I might thing this is some kind of metric or logging tag rather than input to the generator.

conjure-api/src/main/conjure/conjure-api.yml
bmoylan, line 161: I'm just realizing this is on the endpoint and not the arguments. If this is meant to replace the `Safe` marker, don't we need it on every arg?

docs/spec/conjure_definitions.md
carterkozak: > I might thing this is some kind of metric or logging tag rather than input to the generator.

My understanding is that it may be used for that as well

conjure-api/src/main/conjure/conjure-api.yml
carterkozak: Idea is that it could include some kind of SLO information with a dynamic component, so it's not clear that constant values will be terribly useful.

bmoylan:
Also, would appreciate advice on whether to hold off on https://github.com/palantir/conjure-go/pull/138 because this is intended to replace it or whether we plan on keeping the `Safe` marker convention and that is relatively future-proof to move forward.

docs/spec/conjure_definitions.md
bmoylan: Could we maintain a non-exhaustive list of known tags? I don't think we want to end up in a world where different generators declare slightly different APIs for the same concepts

bmoylan:
cc @tabboud

ferozco:
Sorry for letting this drop folks. Hoping to finish this off this week

conjure-api/src/main/conjure/conjure-api.yml
ferozco: Yes, we should add the tags field to parameters as well. Good catch

ferozco: Adding an enum would suggest to folks that there are only a fixed set of valid options for the field instead of goal which is to give developers (and ourselves) more flexibility in the metadata they can attach to a conjure definitions.

conjure-core/src/main/java/com/palantir/conjure/defs/ConjureParserUtils.java
```diff
@@ -412,7 +413,8 @@ private static HttpPath parseHttpPath(
                     .type(original.type().visit(new ConjureTypeParserVisitor(typeResolver)))
                     .paramType(paramType)
                     .docs(original.docs().map(Documentation::of))
-                    .markers(parseMarkers(original.markers(), typeResolver));
+                    .markers(parseMarkers(original.markers(), typeResolver))
+                    .tags(original.tags());
```
carterkozak, line 417: maybe worth validating against empty strings, leading/trailing whitespace? No strong opinion

carterkozak (approved):

sfackler:
https://github.com/palantir/conjure-rust/pull/118

PR #745, opened 2020-11-07
Integrating the list of HTTP methods.

ndrglv:
## Before this PR
Multiple common HTTP methods are missing.

## After this PR
Multiple common HTTP methods have been added.

## Possible downsides?
None

### DIFF
--- a/conjure-api/src/main/conjure/conjure-api.yml
+++ b/conjure-api/src/main/conjure/conjure-api.yml
@@ -165,6 +165,9 @@ types:
       HttpMethod:
         values:
           - GET
+          - HEAD
+          - OPTIONS
+          - PATCH
           - POST
           - PUT
           - DELETE
--- a/docs/concepts.md
+++ b/docs/concepts.md
@@ -4,7 +4,7 @@ _Conjure offers the following abstractions to use when defining your APIs. To se
 
 
 ### HTTP endpoints
-- `GET`, `PUT`, `POST`, `DELETE` - [HTTP methods](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods).
+- `GET`, `HEAD`, `OPTIONS`, `PATCH`, `DELETE`, `POST`, `PUT` - [HTTP methods](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods).
 - _Query parameters_ - e.g. `https://example.com/api/something?foo=bar&baz=2`
 - _Path parameters_ - Parsed sections of URLs e.g. `https://example.com/repo/{owner}/{repo}/pulls/{id}`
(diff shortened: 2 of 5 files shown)


palantirtech:
Thanks for your interest in [palantir/conjure](https://github.com/palantir/conjure), @ndrglv! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

iamdanfox:
Hey @ndrglv, it might seem a bit weird, but conjure is actually _intentionally_ quite restrictive here because we've found that huge graphs of inter-service communication can actually be built up on top of the small number of primitives that are already baked in (GET/POST/PUT/DELETE).

Adding HEAD/OPTIONS/PATCH all have subtle implications, which I'm not sure are net positive.  For example,

- HEAD - people usually make this call when they're just interested in getting headers back (or even just testing their authn/z). Conjure doesn't actually expose the ability to specify arbitrary response headers (again, to try to constrain the different possibly building blocks that people use to construct APIs), so the usefulness of conjure-defined HEAD endpoints seems kinda reduced.
- OPTIONS - our servers do implement this, but we haven't seen a need to allow people to define API endpoints that solely respond to the OPTIONS verb.
- PATCH - this is the most subtle of all. The PATCH verb is generally used to modify one aspect of a resource (with the assumption that the rest is left unchanged).  This is often done in an untyped way, where you can just send the JSON keys you want to modify but leave the rest.  The Conjure API definition language places a strong emphasis on using types everywhere, which seems kinda incompatible with the idea of arbitrary patching.

Interestingly even though the language doesn't allow people to define APIs with HEAD/OPTIONS/PATCH, our [clients](https://github.com/palantir/dialogue/blob/develop/dialogue-target/src/main/java/com/palantir/dialogue/HttpMethod.java) understand how to make these requests and our servers actually do respond to HEAD/OPTIONS ones.

Does this rationale make sense? Perhaps you can flesh out more of the use-case you have, or what you're currently unable to do?

ndrglv:
I agree with your points in principle - however, we are finding the PATCH method in particular is actually widely used in third party services precisely to modify resources, especially async remote jobs. The practical implications of not having the PATCH option is that either we do not use Conjure, or we need to break out all PATCH based parts of a service and then spool up the "normal" service alongside the "patch" service wherever we are using that service, which is not only impractical but also decidedly unelegant.

iamdanfox:
I think it's OK to hand roll a few jersey interfaces if you need to talk to external services. At the end of the day it's a bit of a happy coincidence if a conjure-spec client actually fully works with arbitrary external services 🤷 - the contract we aim for is conjure clients talking to conjure services.

PR #767, opened 2020-12-02
 Fix ForeignRefType namespace parsing not adhering to specs

lhhong:
## Before this PR
Namespace in parsing of ForeignReferenceType would only accept `[a-zA-Z]+` while the specs accepts `^[_a-zA-Z][_a-zA-Z0-9]*$` as per https://github.com/palantir/conjure/blob/master/docs/spec/conjure_definitions.md#typesdefinition

## After this PR

==COMMIT_MSG==
ForeignRefType namespace parsing now adheres to specs, accepting numbers and underscores
==COMMIT_MSG==


palantirtech:
Thanks for your interest in [palantir/conjure](https://github.com/palantir/conjure), @lhhong! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

lhhong:
👋

conjure-core/src/test/java/com/palantir/conjure/parser/types/TypeParserTests.java
```diff
@@ -85,6 +85,9 @@ public void testParser_refType() throws ParseException {
     public void testParser_foreignRefType() throws ParseException {
         assertThat(TypeParser.INSTANCE.parse("bar.Foo"))
                 .isEqualTo(ForeignReferenceType.of(Namespace.of("bar"), TypeName.of("Foo")));
+
+        assertThat(TypeParser.INSTANCE.parse("bar_1.Foo"))
+                .isEqualTo(ForeignReferenceType.of(Namespace.of("bar_1"), TypeName.of("Foo")));
```
fawind, line 90: Can we add a test that covers the `notAllowedToStartWith` codepath e.g. for `1_bar.Foo`?

fawind:
Hey lhhong, sorry for the delay in response, this review slipped through! Makes sense to expand the parser and match the spec. The code looks very clean, would just like to see another test where the parser fails.

conjure-core/src/main/java/com/palantir/conjure/parser/types/TypeParser.java
```diff
@@ -90,8 +90,15 @@ public LocalReferenceType parse(ParserState input) throws ParseException {
...
             }
+            TypeName typeName;
+            try {
+                typeName = TypeName.of(typeReference);
+            } catch (IllegalArgumentException _e) {
+                input.rewind();
+                return null;
+            }
             input.release();
-            return LocalReferenceType.of(TypeName.of(typeReference));
```
lhhong, line 94: The negative test case was getting the wrong exception because the parsing was trapped here, which was how I ran into this bug in the first place, getting a cryptic and totally irrelevant exception.

fawind: Makes sense!

For my own reference of why this is needed in `LocalReferenceTypeParser` and not `ForeignReferenceTypeParser`, this is because we go to the next parser if the previous one returned `null` and `ForeignReferenceTypeParser` is before `LocalReferenceTypeParser`:

https://github.com/palantir/conjure/blob/7ecb0ea62f0717e57b534943ce0dae4ad65c27de/conjure-core/src/main/java/com/palantir/conjure/parser/types/TypeParser.java#L58-L67

fawind:
The changelog bot cannot create a changelog for you because it does not have perms to write to your fork. Can you push up a changelog manually? PR looks good to me!

fawind (approved):

PR #799, opened 2021-02-01
Fix conjure-imports example in conjure_definitions.md 

glynternet:
## Before this PR
The following exception would occur whilst compiling the example.
```
com.palantir.logsafe.exceptions.SafeIllegalArgumentException: Unrecognized definition, types must have either fields, values or an alias defined.
```

## After this PR
When compiling just the example.yml and common.yml there is a NullPointerException, but at least the definition can be parsed properly.


carterkozak (approved):
Thanks!

PR #804, opened 2021-02-16
Update error message of invalid type

wenhoujx:
## Before this PR
```
Caused by: java.lang.IllegalArgumentException: TypeNames must be a primitive type [any, string, integer, double, boolean, safelong, rid, bearertoken, uuid] or match pattern ^[A-Z][a-z0-9]+([A-Z][a-z0-9]+)*$: bytes
``` 
when use invalid type `bytes`, i got that error message on `classpath 'com.palantir.gradle.conjure:gradle-conjure:5.1.1'`. It's missing `binary` and `datetime` 
## After this PR

==COMMIT_MSG==
Includes binary and datetime in suggested built-in types in the error message of invalid type. 
==COMMIT_MSG==

### DIFF
--- a/conjure-core/src/main/java/com/palantir/conjure/parser/types/names/TypeName.java
+++ b/conjure-core/src/main/java/com/palantir/conjure/parser/types/names/TypeName.java
@@ -19,10 +19,13 @@ package com.palantir.conjure.parser.types.names;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonValue;
 import com.google.common.base.Preconditions;
-import com.google.common.collect.ImmutableSet;
 import com.palantir.conjure.defs.ConjureImmutablesStyle;
 import com.palantir.conjure.parser.types.NamedTypesDefinition;
+import com.palantir.conjure.spec.PrimitiveType;
+import java.util.Arrays;
+import java.util.Set;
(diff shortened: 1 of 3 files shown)


conjure-core/src/main/java/com/palantir/conjure/parser/types/names/TypeName.java
```diff
@@ -34,8 +37,11 @@
 public abstract class TypeName {
 
     private static final Pattern CUSTOM_TYPE_PATTERN = Pattern.compile("^[A-Z][a-z0-9]+([A-Z][a-z0-9]+)*$");
-    static final ImmutableSet<String> PRIMITIVE_TYPES =
-            ImmutableSet.of("any", "string", "integer", "double", "boolean", "safelong", "rid", "bearertoken", "uuid");
+    static final Set<String> PRIMITIVE_TYPES = Arrays.stream(PrimitiveType.Value.values())
```
carterkozak, line 40: The values static method is simpler and avoids the UNKNOWN check.
```suggestion
    static final Set<String> PRIMITIVE_TYPES = PrimitiveType.values().stream()<..etc>
```

wenhoujx: 👍

carterkozak (approved):

PR #811, opened 2021-03-16
Fix TypeParserTests

tetigi:
## Before this PR
Based on my understanding, the test `testParser_foreignRefType` should be throwing an `IllegalArgumentException` rather than a `ParseException` on `1_bar.Foo`, as it doesn't match the required type pattern.

Not sure why this is valid in `master` - the `TypeParser` currently eats `IllegalArgumentException`s, which means that `testInvalidNames` does not pass. If I run this `TypeParserTests` locally on `master` it always fails for me.
## After this PR
==COMMIT_MSG==
Fix type parser to throw informative `IllegalArgumentException` instead of `ParseException`.
==COMMIT_MSG==

### DIFF
--- a/conjure-core/src/main/java/com/palantir/conjure/parser/types/TypeParser.java
+++ b/conjure-core/src/main/java/com/palantir/conjure/parser/types/TypeParser.java
@@ -90,13 +90,7 @@ public enum TypeParser implements Parser<ConjureType> {
                 input.rewind();
                 return null;
             }
-            TypeName typeName;
-            try {
-                typeName = TypeName.of(typeReference);
-            } catch (IllegalArgumentException _e) {
-                input.rewind();
-                return null;
-            }
+            TypeName typeName = TypeName.of(typeReference);
             input.release();
             return LocalReferenceType.of(typeName);
         }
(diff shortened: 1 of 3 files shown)


palantirtech:
Thanks for your interest in [palantir/conjure](https://github.com/palantir/conjure), @tetigi! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

conjure-core/src/main/java/com/palantir/conjure/parser/types/TypeParser.java
```diff
@@ -90,13 +90,7 @@ public LocalReferenceType parse(ParserState input) throws ParseException {
                 input.rewind();
                 return null;
             }
-            TypeName typeName;
-            try {
-                typeName = TypeName.of(typeReference);
-            } catch (IllegalArgumentException _e) {
-                input.rewind();
-                return null;
```
carterkozak, line 98: I think the catch here is important in case the input is not expected to be handled by `ForeignReferenceTypeParser` but is handled by `TypeReferenceParser`.
Have you seen https://github.com/palantir/conjure/pull/767/files#r576677144

PR #813, opened 2021-03-16
Checked exceptions on endpoints

tetigi:
## Before this PR

Currently, all declared errors in a Conjure definition are unchecked. They can be called from anywhere, on any endpoint, at any time.
## After this PR

==COMMIT_MSG==
Add the ability to attach errors to endpoint definitions. In Java-land, these could be realised as *checked* exceptions on the endpoint, thus mandating that they be handled by consumers.
==COMMIT_MSG==


carterkozak:
Let's convert this to a draft for now, since we're not planning to consider merging until we have an RFC + example implementations as well.

conjure-api/src/main/conjure/conjure-api.yml
```diff
@@ -159,6 +159,7 @@ types:
           deprecated: optional<Documentation>
           markers: list<Type>
           tags: set<string>
+          errors: list<ErrorDefinition>
```
carterkozak, line 162: I think we may want to use `set<TypeName>` here so that a single error definition may be used across several endpoints. We can validate that all error type-names can be found in the top level error definitions.

robert3005:
I like the idea. I would push for implementation to be the same between all languages such that mental overhead of using conjure would be the same.

carterkozak:
@robert3005 How we surface this across languages depends on what features the language supports to detect failure to handle errors at build time.

tetigi:
Closing in lieu of https://github.com/palantir/conjure/pull/816

PR #814, opened 2021-03-16
Add behaviour to the 'or' parser to pass through errors

tetigi:
## Before this PR

When an exception is hit while performing `Parser.or`, the exception is immediately thrown. Additionally, this current behaviour is causing tests to fail on master.
## After this PR

==COMMIT_MSG==
Change behaviour of `or` parser to instead store any exceptions, and re-throw it as a `ParseException` if no parser successfully completes.
==COMMIT_MSG==

### DIFF
--- a/conjure-core/src/main/java/com/palantir/parsec/Parsers.java
+++ b/conjure-core/src/main/java/com/palantir/parsec/Parsers.java
@@ -153,11 +153,7 @@ public final class Parsers {
             }
 
             if (exception != null) {
-                ParseException toThrow = new ParseException(exception.getMessage(), input);
-                for (Throwable suppressed : exception.getSuppressed()) {
-                    toThrow.addSuppressed(suppressed);
-                }
-                throw toThrow;
+                throw new ParseException(exception.getMessage(), input, exception);
             }
 
             return null;
--- a/conjure-core/src/main/java/com/palantir/parsec/Parsers.java
+++ b/conjure-core/src/main/java/com/palantir/parsec/Parsers.java
@@ -108,6 +108,9 @@ public final class Parsers {
     /**
      * Parse either firstOption or iterate through otherOptions and return result.
      *
+     * Exceptions are caught while trying options - if no options succeed, then the last exception thrown is re-thrown
+     * as a {@code ParseError}.
+     *
      * @param <T> the type the parsers return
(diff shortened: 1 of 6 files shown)


conjure-core/src/main/java/com/palantir/parsec/Parsers.java
```diff
@@ -108,28 +108,47 @@ public T parse(ParserState input) throws ParseException {
...
+            }
+
+            if (result != null) {
+                return result;
+            }
+            for (Parser<? extends T> nextOption : otherOptions) {
+                try {
+                    result = gingerly(nextOption).parse(input);
+                } catch (RuntimeException e) {
+                    exception = e;
```
carterkozak, line 139: Maybe something along these lines
```suggestion
                    if (exception == null) {
                        exception = e;
                    } else {
                        exception.addSuppressed(e);
                    }
```

carterkozak: Do we also need to `input.rewind();` in this case?

tetigi: Oh I wasn't aware of suppressed exceptions - looks great!

tetigi: What does rewinding achieve if we're throwing a `ParseException`?

tetigi: Nevermind - I see.

conjure-core/src/main/java/com/palantir/parsec/Parsers.java
```diff
@@ -108,28 +108,59 @@ public T parse(ParserState input) throws ParseException {
...
                 }
-                for (Parser<? extends T> nextOption : otherOptions) {
-                    result = gingerly(nextOption).parse(input);
-                    if (result != null) {
-                        return result;
-                    }
+            }
+
+            if (exception != null) {
+                ParseException toThrow = new ParseException(exception.getMessage(), input);
```
carterkozak, line 156: Let's add a `ParseException` constructor overload so we don't have to copy over the suppressed values, they're included in the cause. This way we don't lose the stack trace data from the initial exception.

PR #815, opened 2021-03-16
Add behaviour to the 'or' parser to pass through errors

carterkozak:
This is a recreation of #814 without a fork, allowing circle builds to run.

## Before this PR

When an exception is hit while performing `Parser.or`, the exception is immediately thrown. Additionally, this current behaviour is causing tests to fail on master.
## After this PR

==COMMIT_MSG==
Change behaviour of `or` parser to instead store any exceptions, and re-throw it as a `ParseException` if no parser successfully completes.
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-814.v2.yml
+++ b/changelog/@unreleased/pr-814.v2.yml
@@ -1,0 +1,7 @@
+type: fix
+fix:
+  description: >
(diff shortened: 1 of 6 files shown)


tetigi (approved):
Thanks for porting it over - I'll make sure to unfork the others too.

carterkozak:
:+1:

PR #821, opened 2021-03-22
Add RFC 007: Conjure Endpoint Errors

tetigi:

### DIFF
--- a/docs/rfc/007-endpoint-errors.md
+++ b/docs/rfc/007-endpoint-errors.md
@@ -1,0 +1,137 @@
+# [RFC] Conjure Endpoint Errors
+
+**Date:** 2020/03/19
+
+**Stakeholders:**
+* *Anyone that produces or consumes conjure-generated code via Dialogue or Undertow.*
+
+* * *
+
+## **Context**
+
+Currently, errors can be defined in Conjure type definitions. These errors can then be thrown by services as `ServiceException`s (which are `RuntimeException`s), and if a client knows about them, they can be caught during runtime by clients (packaged up as `RemoteException`s).
+
+There are multiple issues with this. As a consumer, you only know that an endpoint may throw an error by reading the documentation or the implementation itself. As a producer, you are limited in the expressivity of what you can say with your exception, as you have no guarantee that a consumer will handle your exception (or even read your documentation).
+
+This RFC introduces *declared errors* on Conjure endpoints. These errors represent real outcomes of a call to an endpoint, and must be handled by consumers.
+
+## Existing Documentation
+
+### PRs
+
+* Root PR (Conjure API changes): https://github.com/palantir/conjure/pull/816
+* Conjure Java API Changes: https://github.com/palantir/conjure-java-runtime-api/pull/634
+* Conjure Java & Dialogue implementation: https://github.com/palantir/conjure-java/pull/1275
+
+## Example
+
+**Conjure API**
+
+```
+types:
+  definitions:
+    default-package: com.palantir.my.api
+    errors:
+      BlueMoon:
+        namespace: MyConjureService
+        code: CUSTOM_SERVER
+
+services:
+  MyConjureService:
+    name: An example service.
+    package: com.palantir.my.api
+    default-auth: header
+    endpoints:
+      getSomeResult:
+        http: GET /
+        returns: list<rid>
+        errors:
+          - BlueMoon
+        docs: Returns some result, but once in a blue moon throws an error
+```
+
+
+**Undertow Service**
+
+```
+@Generated("com.palantir.conjure.java.services.UndertowServiceInterfaceGenerator")
+public interface MyConjureService {
+    /**
+     * Returns some result, but once in a blue moon throws an error
+     * @apiNote {@code GET /}
+     */
+    List<ResourceIdentifier> getSomeResult(AuthHeader authHeader)
+            throws MyConjureServiceErrors.BlueMoonServiceException;
+```
+
+**Undertow Resource**
+
+```
+public final class MyConjureResource implements MyConjureService {
+
+   ...
+
+   @Override
+   public List<ResourceIdentifier> getSomeResult(AuthHeader authHeader)
+          throws BlueMoonServiceException {
+      throw MyConjureServiceErrors.throwIfBlueMoon(isBlueMoon);
+   }
+```
+
+
+**Dialogue Client**
+
+```
+MyConjureServiceBlocking myConjureService = 
+    witchcraft.conjureClients()
+        .client(MyConjureServiceBlocking.class, "my-conjure-service").get();
+
+try {
+    myConjureService.getSomeResult(AuthHeader.valueOf("authHeader"));
+} catch (BlueMoonRemoteException e) { // <- checked exception so has to be caught!
+    handleBlueMoonEvent();
+}
+```
+
+
+
+## Proposal
+
+This proposal should be viewed in two distinct parts - declared (and handled) exceptions on Conjure endpoints, and the language implementations of such a feature.
+
+### Intent
+
+The primary intent of this proposal is to provide more expressivity around errors in a Conjure API, while minimizing the change surface-area, especially around wire format changes. Our proposed change introduces *no* wire changes, so should be purely additive and not dramatically affect back-compat.
+
+### Declared Errors
+
+**Semantics:** To declare an error on an endpoint is to say “This is a very real outcome of calling this endpoint. You should handle this outcome to ensure that you can react accordingly”. It should be something closely tied to the functionality of that endpoint - that is to say, `getFoo` declaring `FooNotReadyYet` would be a good example, but `UserNotAuthorized` would be undeclared / passed through as normal (as it is a standard expectation of interacting with an authorized service).
+
+**Pros:**
+
+* Consumers now know the *expected* outcomes of calling an endpoint in a first-class way and can handle those outcomes accordingly.
+* Producers are able to express the different failure states of their endpoints in greater detail, and have conviction that those failure states are being handled by consumers.
+* Changing the failure states of an endpoint now visibly affects the API definition - changes in behavior are properly captured in code (rather than just documentation).
+
+**Cons:**
+
+* To be effective, these failure cases will need to be ‘forced’ upon consumers via the compiler. This could be as complex data types (perhaps unions of {success, error1, error2, ...} or via checked exceptions. This would entail some degree of (justified) work migrating to a with-errors world to wire each failure case into the correct behavior in client code.
+* Given the new flexibility, producers may more eagerly throw errors in their code, and ‘old’ clients will have a hard time dealing with these without upgrading.
+* Producers might be incentivized to cram *all* the possible outcomes of their endpoint as errors (see the `UserNotAuthorized` example above), and as such unduly burden consumers of their endpoint with things that should probably just be `RuntimeException`s.
+* Introducing further types (whether as checked exceptions or otherwise) that determine control flow may result in them spreading throughout a code base and result in further coupling to the RPC layer.
+
+### Implementation
+
+Declaring errors on endpoints not only allows producers to enforce that each error case is going to be handled by a consumer, but it also allows us to provide the tools to make that ergonomic for consumers. Exactly how those ergonomics will be provided depend on the language. There were two main approaches that were considered:
+
+**Enriched Types**
+
+With an expressive type system, it is possible to express the different possible outcomes as an enriched version of the normal return type. This could be done, for example, by using union types, which already exist in the Conjure universe. Some languages are more amiable to union types than others - even now Java doesn’t really provide the necessary tools to do this in a succinct way (ie. sealed classes with switch’ing over instances such as [Sealed classes | Kotlin](https://kotlinlang.org/docs/sealed-classes.html#sealed-classes-and-when-expression)).
+
+**Checked Exceptions**
+
+Java provides checked exceptions as a method of enforcing that exceptional states are handled, and declared errors on endpoints fit nicely into that.
+
+**Why We Chose Checked Exceptions**
+
+As of now, `SerializableError` does not lend itself to complex data types as complex error objects are flattened into a `Map<String, String>`, which makes parsing it out very difficult. Checked exceptions preserve the existing status-quo (that is, not messing with the reality of  `SerializableError`), but still improve expressivity and enforce errors being actually handled. Additionally, they are just more ergonomic in our current version of Java.


docs/rfc/007-endpoint-errors.md
```diff
@@ -0,0 +1,137 @@
+# [RFC] Conjure Endpoint Errors
+
+**Date:** 2020/03/19
+
+**Stakeholders:**
+* *Anyone that produces or consumes conjure-generated code via Dialogue or Undertow.*
```
carterkozak, line 6: It impacts everyone, although it may not be a requirement that every implementation adds generator support
```suggestion
* *Anyone that produces or consumes conjure-generated code.*
```

docs/rfc/007-endpoint-errors.md
```diff
@@ -0,0 +1,137 @@
...
+
+* * *
+
+## **Context**
+
+Currently, errors can be defined in Conjure type definitions. These errors can then be thrown by services as `ServiceException`s (which are `RuntimeException`s), and if a client knows about them, they can be caught during runtime by clients (packaged up as `RemoteException`s).
```
carterkozak, line 12: Let's refactor this a bit to discuss the protocol, with a separate section describing how it impacts existing implementations (java).

docs/rfc/007-endpoint-errors.md
```diff
@@ -0,0 +1,229 @@
...
+myConjureService.getSomeResult(AuthHeader.valueOf("authHeader"), MyConjureServiceErrors.getSomeResultBuilder()
+    .blueMoon(e -> handleBlueMoonEvent())
+    .ok(mySafeResult3 -> handleOk(mySafeResult3))
+    .build());
+    
+// v- or via `Visitor<T>` and return `T` in all branches
+SomeOtherResult otherResult = myConjureService.getSomeResult(AuthHeader.valueOf("authHeader"), MyConjureServiceErrors.getSomeResultBuilder()
+    .blueMoon(e -> handleBlueMoonEvent())
+    .ok(mySafeResult3 -> handleOk(mySafeResult3))
+    .build());
```
carterkozak, line 132: This style visitor was used as an example in the original discussion for why folks did not want to simply return a union from their endpoint.

docs/rfc/007-endpoint-errors.md
```diff
@@ -0,0 +1,229 @@
...
+### Mandatory vs. Optional
+
+As discussed above, declared errors can be either made to be optional, or mandatory. Both versions come with their own
+pros and cons, and picking one side or the other inevitably involves making sacrifices.
+
+**The case for mandatory handling**
+
+_Forcing_ consumers to handle your exceptional cases is beneficial to the _producer_. You can have confidence that
+consumers will be encouraged to handle the exceptional cases, even if they are not reading your documentation. Acknowledging
+the presence of exceptional cases will be required at the minimum.
```
carterkozak, line 184: > Acknowledging the presence of exceptional cases will be required at the minimum.

This is the goal based on the second paragraph

> As a consumer, you only know that an endpoint may throw an error by reading the 
documentation or the implementation itself. As a producer, you are limited in the expressivity of what you can say 
with your exception, as you have no guarantee that a consumer will handle your exception (or even read your documentation).

docs/rfc/007-endpoint-errors.md
```diff
@@ -0,0 +1,229 @@
...
+## **Context**
+
+Currently, errors can be defined in Conjure type definitions. These errors can then be thrown by services, and if a client 
+knows about them, they can be handled during runtime by clients. The exact handling of this mechanism may vary by
+the implementation (for example, in Java these are `ServicException`s, which become `RemoteException`s which can be caught
+by the client).
+
+There are multiple issues with this. As a consumer, you only know that an endpoint may throw an error by reading the 
+documentation or the implementation itself. As a producer, you are limited in the expressivity of what you can say 
+with your exception, as you have no guarantee that a consumer will handle your exception (or even read your documentation).
```
iamdanfox, line 19: as a bystander here, I'm interested to hear a bit more about the context/motivation here... were there some specific internal cases that motivated this? or is it a broader hope that by defining a new first-class way of declaring errors in conjure that people will use them more?

docs/rfc/007-endpoint-errors.md
```diff
@@ -0,0 +1,229 @@
...
+  MyConjureService:
+    name: An example service.
+    package: com.palantir.my.api
+    default-auth: header
+    endpoints:
+      getSomeResult:
+        http: GET /
+        returns: list<rid>
+        errors:
+          - BlueMoon
```
iamdanfox, line 55: What assumptions about completeness/exhaustiveness should users be able to make here?

Clearly in Java, runtime exceptions are _much_ more ergonomic to use (i.e. in streams/lambdas), so it seems like people will not adopt checked exceptions _everywhere_ in backends.

In that case, I worry that someone will introduce a new conjure error type and start throwing it deep inside one of their stores or implementation detail classes, not realizing that there are actually _multiple_ endpoints which call this and could now start throwing these errors?

carterkozak: > In that case, I worry that someone will introduce a new conjure error type and start throwing it deep inside one of their stores or implementation detail classes, not realizing that there are actually multiple endpoints which call this and could now start throwing these errors?

Checked exceptions are validated by the type system, attempting to throw a checked exception from an endpoint that doesn't declare one will fail to compile. I'd expect this to result in the api definition being updated to better reflect possible failures for each endpoint. You're correct that the exception may be wrapped with an unexpected runtime exception, but isn't that true of runtime exceptions as well?

docs/rfc/007-endpoint-errors.md
tetigi: The wording may be unclear - the sentence is probably better read as 'Through mandatory exceptions, users will at the very minimum have to acknowledge the existence of your error cases and consciously decide to not handle them'.

Is that what you were referring to?

docs/rfc/007-endpoint-errors.md
tetigi: There are some examples that have been experienced in the wild. As an example:

- Consumers of an API assume there are no limits to what can be returned from a batch endpoint.
- They write code that interacts with this.
- At some point they reach a sufficient scale where the batch endpoint hits the limit, and it starts erroring.
- This can result in parts of a UI or service stopping working, and bad times for everyone.

If that endpoint had had a way to say 'hey, btw there is a very real chance of a bad edge case occurring' then they might have programmed defensively against that (of course, in this case they probably should have used pages or something similar too).

I think the idea behind declared errors is to have a generalised way to embed bad edge cases into your API definition.

docs/rfc/007-endpoint-errors.md
tetigi: My intention here was purely as the 'shell' around the existing error mechanism - so we still ship the same errors across the wire (so the JSON is identical to both existing and the exceptions-based mechanism) but we just shepherd the result into a union rather than throwing it as an exception.

docs/rfc/007-endpoint-errors.md
raiju: Note that the generator doesn't need to use checked exceptions. We could generate `Result<...>` types which are very stream/lambda friendly.

ferozco: One issue that's come up with Result types in Java code is that developers have to remember to inspect the result and convert the error state to an exception. Not a deal breaker, (and something we can handle with errorprone) but worth noting

docs/rfc/007-endpoint-errors.md
```diff
@@ -0,0 +1,229 @@
...
+There are multiple issues with this. As a consumer, you only know that an endpoint may throw an error by reading the 
+documentation or the implementation itself. As a producer, you are limited in the expressivity of what you can say 
+with your exception, as you have no guarantee that a consumer will handle your exception (or even read your documentation).
+
+This RFC introduces *declared errors* on Conjure endpoints. These errors represent real outcomes of a call to an endpoint, 
+and should be handled by consumers.
+
+## Existing Documentation
+
+### PRs for example implementation (in Java, using checked exceptions)
```
ferozco, line 26: Have you guys thought about how declared errors would be consumed in other languages? It would be interesting to see an idea of what the implementation for another commonly used language (TypeScript, Golang) would look like.

docs/rfc/007-endpoint-errors.md
```diff
@@ -0,0 +1,229 @@
...
+   ...
+
+   @Override
+   public List<ResourceIdentifier> getSomeResult(AuthHeader authHeader)
+          throws BlueMoonServiceException {
+      throw MyConjureServiceErrors.throwIfBlueMoon(isBlueMoon);
+   }
+```
+
+### Example Dialogue Client Implementation via Exceptions
```
ferozco, line 87: Its worth noting that checked exceptions would not work with async clients. Have we thought about how we would implement declared errors in this case? It just seems like a hard drop in safety if you move to async clients

docs/rfc/007-endpoint-errors.md
raiju: I'd prefer returning a result type here, to allow for scoped handling:

```
// Lazy
myConjureService.getSomeResult(AuthHeader.valueOf("authHeader")).orThrow();

// Cleaner
myConjureService.getSomeResult(AuthHeader.valueOf("authHeader")).handle(success => ..., failure => ...);

// Stream which drops failures
myList.stream()
    .map(entry => myConjureService.getSomeResult(AuthHeader.valueOf("authHeader"), entry))
    .flatMap(Result::successful)
    .collect(Collectors.toList())

// Stream which handles failures
myList.stream()
    .map(entry => myConjureService.getSomeResult(AuthHeader.valueOf("authHeader"), entry))
    .map(result => result.or(MyConjureServiceErrors.getSomeResultBuilder()
        .blueMoon(e -> deriveBackupValue(e))
        .build()))
    .collect(Collectors.toList())

```

docs/rfc/007-endpoint-errors.md
```diff
@@ -0,0 +1,229 @@
...
+the presence of exceptional cases will be required at the minimum.
+
+There are benefits as a consumer as well - additional errors, behaviors, or edge cases will be exposed to you on upgrading
+your client, and you should (in theory) never miss out on a new case (assuming you keep your client relatively up to date).
+
+**The case for optional handling**
+
+Optional handling favors the _consumer_ - existing code and behaviors will not have to change at the outset, and they
+can opt into improved error handling as necessary. The obvious downside is that producers could publish new errors until
+the cows come home, but lazy consumers may never actually handle these edge cases.
```
ferozco, line 193: I think the downside is a little more subtle here. There is still value for producers defining more specific and rich errors since it helps clients better understand what happened. The cost is that clients will have a harder time _proactively_ handling error states

docs/rfc/007-endpoint-errors.md
carterkozak: > Its worth noting that checked exceptions would not work with async clients.

I'd expect the return type to become more complex than the existing ListenableFuture. The future is already a form of special `Result` object. Async clients are already used much less broadly than other clients and much easier to misuse. That's not to say we should make it easier to misuse, but I wouldn't use it as an argument against adding safety to most common use cases.

docs/rfc/007-endpoint-errors.md
```diff
@@ -0,0 +1,229 @@
...
+pros and cons, and picking one side or the other inevitably involves making sacrifices.
+
+**The case for mandatory handling**
+
+_Forcing_ consumers to handle your exceptional cases is beneficial to the _producer_. You can have confidence that
+consumers will be encouraged to handle the exceptional cases, even if they are not reading your documentation. Acknowledging
+the presence of exceptional cases will be required at the minimum.
+
+There are benefits as a consumer as well - additional errors, behaviors, or edge cases will be exposed to you on upgrading
+your client, and you should (in theory) never miss out on a new case (assuming you keep your client relatively up to date).
```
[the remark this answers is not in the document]
carterkozak: The way I've mentally modeled endpoint-errors is as a way to recommend that clients handle particular types of errors, not necessarily to enumerate all possible types of failures. How would you use the data from `endpoint xyz threw an undeclared internal precondition error`?
In some cases we don't expect a client to handle a failure, just to fail. I don't think we would want to force callers to enumerate cases that are inconsequential or equivalent. UNAUTHORIZED and FORBIDDEN ErrorCodes are remarkably common, and can be expected from any authed endpoint.

docs/rfc/007-endpoint-errors.md
bavardage: We can't make any claims about exhaustiveness because there's no guarantee that people stay up to date with the generated client bindings.

docs/rfc/007-endpoint-errors.md
```diff
@@ -0,0 +1,229 @@
...
+
+**Pros:**
+
+* Consumers now know the *expected* outcomes of calling an endpoint in a first-class way and can handle those outcomes accordingly.
+* Producers are able to express the different failure states of their endpoints in greater detail, and have conviction that those failure states are being handled by consumers.
+* Changing the failure states of an endpoint now visibly affects the API definition - changes in behavior are properly captured in code (rather than just documentation).
+
+**Cons:**
+
+* To be effective, these failure cases will need to be ‘forced’ upon consumers via the compiler. This could be as complex
```
bavardage, line 162: and thus also requires people take regular upgrades of generated api bindings

docs/rfc/007-endpoint-errors.md
```diff
@@ -0,0 +1,229 @@
...
+* Changing the failure states of an endpoint now visibly affects the API definition - changes in behavior are properly captured in code (rather than just documentation).
+
+**Cons:**
+
+* To be effective, these failure cases will need to be ‘forced’ upon consumers via the compiler. This could be as complex 
+  data types (perhaps unions of {success, error1, error2, ...} or via checked exceptions. This would entail some degree 
+  of (justified) work migrating to a with-errors world to wire each failure case into the correct behavior in client code.
+* Given the new flexibility, producers may more eagerly throw errors in their code, and ‘old’ clients will have a hard 
+  time dealing with these without upgrading.
+* Producers might be incentivized to cram *all* the possible outcomes of their endpoint as errors (see the
```
bavardage, line 167: yeah - how do we capture that the intent is more result-type-esque (imo) vs listing all of the truly 'exceptional' cases

docs/rfc/007-endpoint-errors.md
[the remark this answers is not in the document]
bavardage: > assuming you keep your client relatively up to date
do we have any data on how this looks in reality? I don't think we generally upgrade -api  dependencies, do we? (this may have changed)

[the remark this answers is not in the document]
carterkozak: We do! Between the minimum-version calculation and pdeps-bot checks we've been able to start excavating api upgrades that trail releases by ~1 month by default.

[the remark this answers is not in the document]
ferozco: Internally we have an excavator that tried to keep clients 45 days behind the latest release

docs/rfc/007-endpoint-errors.md
```diff
@@ -0,0 +1,229 @@
...
+actually handled. 
+
+## Open Questions ##
+
+1) Should handling errors on endpoints be mandatory? While it seems like a good idea in principle, could it cause 'bad'
+   behavior in client or server code?
+2) In the Java implementation, are checked exceptions the way to go? There are several other possible implementations
+   and ideas that may be more in-line with other languages, such as using a union type via a visitor pattern (which 
+   can be made much less verbose via a 'Visitor Builder').
+3) How important is implementation consistency across languages? Would it be acceptable to have union types in TS, but
```
bavardage, line 226: do we have a stance on this for conjure generally?

My inclination is to opt for whatever feels most natural in each language (e.g. I think in rust we model conjure unions as enums).

docs/rfc/007-endpoint-errors.md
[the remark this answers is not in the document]
bavardage: so does this turn into also then people declare an error in their api, but don't feel ok with actually throwing it until `45 days + median service deploy time` have passed?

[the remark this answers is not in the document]
carterkozak: Not quite. By leveraging the existing conjure error contract, existing clients will handle errors from new servers exactly the same way they do today. The wire format is identical, and all clients support parsing arbitrary conjure errors. Clients using sufficiently new generated code are able to advertise certain types of errors so that developers are inclined to handle them reasonably, and (better yet) avoid causing them in the first place.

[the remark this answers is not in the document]
raiju: If you're adding state that must be parsed client-side to be correct there's only one option: make a new endpoint and add it to the return value. Outside of those, any solution will have its own set of downsides, where we can decide where those downsides land.

This feels like a reasonable trade-off, where we have a mechanism of rolling it out/making sure dev teams can realistically handle known errors, which should also make things like propagating human understandable error causes significantly easier. It does rely on some judgement on the API side on what constitutes an acceptable break and what doesn't.

No solution will be able to make clients forward compatible, nor should it try.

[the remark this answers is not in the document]
raiju: On undeclared errors, the "how would you use the data" for me would be to be able to track which endpoints throw which errors. For errors thrown e.g. closer to the DB layer, I could accidentally introduce an error that I wanted to declare, but forgot to add to all endpoints.

uschi2000:
I would be inclined to split this RFC in 2:
    * First, force service to declare all ServiceExceptions they throw as (checked) Conjure errors in their API defs. This would improve our API documentation and provide a complete list of error states for each endpoint.
    * [maybe later] Second, expose checked exceptions in clients. I wonder if we should make this opt-in? Do we need client-side codegen before we can really do this?

tetigi:
I have restructured the RFC to try to emphasize the split between the actual RFC itself (that is, declared errors as a thing and the motivations / benefits of it), and the implementations.

Reading between the lines, it seems that people are broadly in agreement that actually doing something with the errors in Conjure (and thus having more clarity and what errors can be returned and where they cannot returned) is a good thing, and most of the discussion seems to be around implementation details (checked exceptions vs. result-types) and ensuring that the change is not abused / used incorrectly by service writers.

The second can hopefully be solved by good documentation around the new feature, along with examples of what would and would not be a good candidate for a declared error.

As for the first, we still think that checked exceptions are the way to go for the Java-land implementation of the feature. It is the most obvious implementation that has first-class support within the language, and while it does not directly play with functional niceness and streams, it is not difficult to wrap a client call to abstract the error handling portion and use that within streams if necessary. I have added some wording clarifying these pros and cons into the Java implementation portion of the RFC, as well as stripping out some of the other examples to focus the span of topics covered a bit.

tetigi:
Putting a last call out for further feedback on the above - it'd be great to get any further thoughts / suggestions before EOD tomorrow so that we can start moving forwards with the less contentious parts (ie. the conjure change itself, plus the backend Java implementation).

robert3005:
So I wonder if we should follow the precedent here and use the same logic we had for unions. Unions have a nicer implementation in java than what we implemented but we went with verbose and cross language approach. Wonder what's the reasoning to not follow that pattern here.

IMHO the non contentious pieces are implementations that don't allow for checked exceptions, i.e. typescript or rust.

ferozco:
I think its fair to start the process of updating the IR to include declared errors. In terms of code-gen changes, what if we start by making changes so that clients can access specific remote exceptions in a typed way.

I think part of the problem with the current world in Java is that even if you do choose to catch a `RemoteException` you're forced to deal with the obtuse `SerializableError` which exposes parameters as `Map<String, String>`. Perhaps we can copy from other implementations (ex. [TypeScript](https://github.com/palantir/conjure-typescript/blob/develop/src/commands/generate/__tests__/resources/errors/importError.ts)) and expose a way to convert from a `SerializableError` to a Union over the known errors for a given error namespace. 

ex:
```java

try {
  myClient.doStuff(auth);
} catch (RemoteException e) {
  // all errors within the MyClient namespace
  MyClientError error = MyClientErrors.of(e.getError());
  error.accept(MyClientErrors.Visistor.builder()
    .invalidStuff(err -> ...)
    .unableToDoStuff(err -> ...)
    .unknown((SerializableError error) -> ...)
    .build());
}
```

To me this seems like no regrets work since even in a world with checked exceptions callers may want to handle errors that were not considered "important" by the server. It is also easy to extend such an implementation to work with declared errors (in lieu of checked exceptions) by generating per endpoint unions where the variants are the declared errors.

ex:
```java
try {
  myClient.doStuff(auth);
} catch (RemoteException e) {
  // declared errors for the specific endpoint
  MyClientDoStuffError error = MyClientDoStuffErrors.of(e.getError());
  error.accept(MyClientErrors.Visistor.builder()
    .invalidStuff(err -> ...)
    .unknown((SerializableError error) -> ...)
    .build());
}
```

docs/rfc/007-endpoint-errors.md
pt-achang: I'd personally vote against result types, as they aren't idiomatic java. Basically, I think it would be nice if client bindings for each language could approximate idiomatic error handling in that language, e.g.:

* Java - checked (or unchecked) exceptions
* C#/python - unchecked exceptions
* golang/rust - (result, error) tuples w/ discriminated unions on the error type

docs/rfc/007-endpoint-errors.md
```diff
@@ -0,0 +1,212 @@
...
+    myConjureService.getSomeResult(AuthHeader.valueOf("authHeader"));
+} catch (BlueMoonRemoteException e) { // <- checked exception so has to be caught!
+    handleBlueMoonEvent();
+}
+```
+
+## Open Questions ##
+
+1) Should handling errors on endpoints be mandatory? While it seems like a good idea in principle, could it cause 'bad'
+   behavior in client or server code?
```
pt-achang, line 205: If I'm reading this correctly, this feels like a java-specific question around checked vs. unchecked exceptions. There's no idiomatic equivalent to a checked exception in C# or python (i.e., all errors "optional" in those languages). On the other hand, in languages like golang or rust, all error handling is "mandatory" by design. Basically, I'm not sure the mandatory vs optional question makes sense it the broader ecosystem of other languages.

docs/rfc/007-endpoint-errors.md
pt-achang: I would vote for trying to keep things idiomatic in each potential target language.

docs/rfc/007-endpoint-errors.md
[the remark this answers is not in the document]
j-baker: For this, I think we're likely just not using the right vocabulary to describe this problem. If an endpoint defines a new endpoint that may be thrown, there are probably a few cases that this can happen. I've probably missed a few.

For 5xx returns:
1. I'm switching from a generic server error to a specific server error. Generally, expect you shouldn't be handling 5xx returns.

For 4xx returns
1. I'm switching from a generic client error to a specific client error. This is clearly not API in this model, clients needn't take an action because they shouldn't have been depending on that behaviour.
2. I used to throw one client error, and am changing to throw a different one (e.g. NotFound to DatasetNotFound).
3. I added a new argument to my method, and if you use it improperly you get a MyNewError. Old consumers probably needn't care about this.

The only cases we probably should care about for the evolution component is 2. This is clearly an API break, so conjure backcompat should not let you add that extra declared argument. Stuff around pdeps bot is pretty immaterial - we have an API contract that says that old code should not stop working, we need to maintain that. The encouragement that we can make changes after 45 days is only reasonable once we bump the major version every 45 days - in absence we'll just introduce a new class of weird errors that will happen regularly that the system used to but no longer protects against.

One way around this is to let the server throw an error which explicitly has a 'parent' error, letting one throw the new error but indicate that it used to be an old error. Therefore if the client does not know about the new error, they can process according to the old code.

docs/rfc/007-endpoint-errors.md
```diff
@@ -0,0 +1,212 @@
...
+This proposal should be viewed in two distinct parts - declared (and handled) exceptions on Conjure endpoints, and the 
+language implementations of such a feature.
+
+### Intent
+
+The primary intent of this proposal is to provide more expressivity around errors in a Conjure API, while minimizing the 
+change surface-area, especially around wire format changes. Our proposed change introduces *no* wire changes, so should 
+be purely additive and not dramatically affect back-compat.
+
+### Declared Errors
```
j-baker, line 69: want this RFC to contain information about how this fits in with backwards compatibility guarantees

docs/rfc/007-endpoint-errors.md
robert3005: As a person who deals with many repos I personally prefer consistency over what feels natural in a language. The overhead of “unnatural” isn’t that high but overhead of having to change your mental model when you look at different repo is high in comparison from experience.

docs/rfc/007-endpoint-errors.md
tetigi: @j-baker This RFC should not involve any backwards breaking changes - old clients will continue to throw exceptions in exactly the same way they do currently (as the wire format will not change at all).

I'll add a section on compat to emphasise this.

tetigi:
On further discussion, a few points have become quite clear:

- Checked exceptions may cause more headaches than they solve. Other proposed solutions (such as exception handlers / visitors) may be more ergonomic and fitting for existing code bases, whilst also being able to leverage the compiler to help handle each error. @ferozco has a good example of this above with the `MyClientErrors.Visitor.builder()` pattern.
- There is (justified) confusion over if this will have compat issues. 

I have revised the RFC accordingly, adding a section on why this RFC should not introduce any back or forwards compat issues, and also leaning less heavily into checked exceptions.

Broadly, people seem to agree on the idea of declaring errors in general. This leaves us in a fairly nice place for this RFC, where at the very least we end up with better documentation around error behaviors on endpoints, and can then work on language-specific implementations of that as we go (and direct conversation around that to those repos).

docs/rfc/007-endpoint-errors.md
```diff
@@ -121,72 +121,64 @@ Generally, the client-side implementation should be done using whatever language
...
+        .build();
+}
 
-   @Override
-   public List<ResourceIdentifier> getSomeResult(AuthHeader authHeader)
-          throws BlueMoonServiceException {
-      throw MyConjureServiceErrors.throwIfBlueMoon(isBlueMoon);
-   }
+// alternatively, we could even provide an extra version of the `getSomeResult` method:
+    SomeResult result = myConjureService.getSomeResult(AuthHeader.valueOf("authHeader"),
```
tpetracca, line 174:
```suggestion
    SomeResult result = myConjureService.getSomeResult(
    		AuthHeader.valueOf("authHeader"),
```

tpetracca: nit to make it match our standard code-style as I ended up doing a double-take

docs/rfc/007-endpoint-errors.md
```diff
@@ -0,0 +1,201 @@
...
+and clients that do not upgrade can continue to catch `SerializableError`s as they did before.
+
+The intention instead is to better leverage Conjure types to provide better tooling for categorising errors into their
+types when they are received, and to improve documentation around what errors can be thrown by an endpoint.
+
+#### Backwards Compatibility of Error Fields
+
+As part of upgrading errors to a first-class part of the API, we should additionally treat errors in a first-class way
+when it comes to backwards compatibility. As the new suggested tooling will make errors more usable, accessing fields
+will also be more important, and (just as with regular fields) removing them will become more problematic.
```
carterkozak, line 136: This gets into some of the issues we original described as out of scope, in particular all parameters are currently to-stringed by the java server implementations.

I agree that error data should be backwards compatible in the long term, but for the sake of this RFC I'm not confident we need to specify this. We may need to relax the constraint in order to support proper structured parameters in the future. I'd considered making all java a generated accessors for error fields optional such that deserialization may fail on individual fields without impacting the overall meaning of the error.

docs/rfc/007-endpoint-errors.md
ericanderson: From a Typescript/Javascript perspective, only unchecked exceptions really make sense as exception handling is very rare. 

I would want to get some feedback from other FE devs but I would probably suggest revving major version of generated libraries if we were to do this so I could write code like:

```ts
const r = service. getSomeResult();
if (r.result) {
  doStuffWithListOfRids(r.result);
} else {
  // treat r.exception as a conjure union of exceptions so unknown can be handled.
}

**Review by ericanderson — COMMENTED** [review]
I think this RFC should actively describe the proposed outcome for our primary language bindings.

File: docs/rfc/007-endpoint-errors.md (hunk 15 above)
Comment (ericanderson) [reply docs/rfc/007-endpoint-errors.md:68]: It has always been the case that you cannot break APIs semantically and this is still  true  with the proposed change, however it may encourage people to "fix" their APIs by adding declared exceptions after the fact. This exact problem was alluded to in previous discussions but to highlight it here:

Suppose you had a `TryAgainLaterException` that you threw when you were being throttled. As a client you could have handled this nicely and just having this exception converted to declared doesn't have to be a big deal for client authors. However, now that there is a shiny new feature, a dev decides to add a `ThrottledException` that includes how long you should wait. This change would have always been a semantic break but now we have gone from a world where exceptions were by and large ignored to one where we are saying "Look Here. Declare your exceptions! It is backwards and forwards safe!" when the reality isn't quite so.

Comment (carterkozak) [reply docs/rfc/007-endpoint-errors.md:68]: @ericanderson As you point out, changing the type of an error is a semantic break, and always has been. Your fear is that adding a feature will make developers believe it's safe to make this kind of semantic break, that could be the case, however there are a couple ways of looking at it:
1. The proposal would make error usages more obvious in API review, thus reviewed more closely to avoid semantic breaks in ways that have always been easy to miss in code review.
2. The existing error type may be declared (improving endpoint documentation in a way that may allow better codegen, but no wire or semantic break)

In terms of backwards compatibility:
* Renaming an existing error is a break. This prevents existing clients from effectively handling a failure case that previously worked as expected.
* Removing an error _may_ not be a break. If an error type cannot be thrown anymore (inputs that were not allowed at one point are now allowed) it's reasonable to remove the error. Old clients which handled the failure won't execute error handling code because the server no longer produces that error. However if the same scenario results in a different kind of error, this is a break as described by the first bullet.
* Adding a new error is not a break. This assumes the error was not previously considered some other kind of error that clients may have been expected to handle. When the server may fail in a new way, clients may be asked to handle it. There's an expectation that the new failure cannot _require_ special handling from clients because old clients must continue to work, however this can be leveraged to provide better user-facing messages in the UI or to describe a failure that was previously surfaced as a 500 response that didn't provide sufficient context.

Note that removing an error may not be a break, but if another error is added in the future for an equivalent case but using a different name, we've created the "Renaming an existing error is a break." case.

**Comment by carterkozak** [comment]
I want to highlight this comment, which we should also add into the rfc text itself: https://github.com/palantir/conjure/pull/821#discussion_r599625795

By forcing clients to acknowledge certain, specific failure outcomes, we can either prevent them from occurring in the first place, or provide more actionable information to users.

The case which prompted this RFC discussion revolved around batch endpoints being called with too many inputs. If devs were aware that there was an upper limit, they wouldn't have passed an unbounded array of IDs to the call, and we would have avoided user facing failures.

I don't want to dive into bike-shedding language specific implementation details before we've agreed on the outcomes we're tackling, but from my perspective this feature is worthwhile because it can be implemented in a way that educates api consumers about the endpoint they're interacting with in a way that ensures they don't misuse it -- this means we're avoiding failures proactively rather than reactively. @ferozco the proposal you've outlined in https://github.com/palantir/conjure/pull/821#issuecomment-810572069 is indeed very low regret, however I'm not sure it provides us much utility outside of test code because it doesn't improve our posture against this class of failures we've observed.

**Comment by carterkozak** [comment]
> So I wonder if we should follow the precedent here and use the same logic we had for unions. Unions have a nicer implementation in java than what we implemented but we went with verbose and cross language approach. Wonder what's the reasoning to not follow that pattern here.
> 
> IMHO the non contentious pieces are implementations that don't allow for checked exceptions, i.e. typescript or rust.

@robert3005 my recollection is that we specifically wanted to make unions hard to misuse as APIs evolve (new variants require visitors to be updated) not that we wanted to implement them such that you can write the same code in rust and java. I'm much more concerned with making code legible, obvious, and correct than structuring bindings the same way across languages. If we're not taking advantage of language features, we also need to invest in building and maintaining static analysis rules and IDE plugins to make up the difference.
We had discussions about making the union _wire format_ into something that was easy to represent across supported languages, but that doesn't apply to this discussion because we're not discussing changes to the wire format.

File: docs/rfc/007-endpoint-errors.md (hunk 14 above)
Comment (pt-achang) [reply docs/rfc/007-endpoint-errors.md]: I vote that language ergonomics / idiomatic-ness is more important than trying to force every language to adopt a uniform error-checking mechanism (e.g., a result type packed within a union type that supports errors).

Comment (uschi2000) [reply docs/rfc/007-endpoint-errors.md]: +1 to @pt-achang 's point

File: docs/rfc/007-endpoint-errors.md (hunk 5 above)
Comment (iamdanfox) [reply docs/rfc/007-endpoint-errors.md:19]: For any other readers I think the specific RFC we're talking about here is titled `[RFC] Handling limits of batch endpoints [2021-03-01]`. I am **very concerned** about the idea of pushing for a `LimitExceededError` to solve this specific case, as I worry it would lead to every client re-implementing the same crap to detect limit exceeded and then do some funky defensive "try again with fewer ids" stuff.

```
batchGetEmployees:
  http: POST ...
  args:
    pageSize:
      param-type: query
      type: optional<string>
    pageToken:
      param-type: query
      type: string
    request: BatchGetEmployeesRequest
  returns: BatchGetEmployeesResponse
  errors:
    - LimitExceededError # <- proposed addition
```

My preferred solution for that RFC is already mentioned lower down: `Implement batching via paging patterns` . This is the way that all other servers I know of handle this, and crucially it can be done transparently on the server side so that all clients benefit without having to write/maintain any funky code.

I think my request here is for a few more _concrete, idiomatic cases_ where this feature would be immediately adopted, because if it's all motivated by this LimitExceeded thing and hypotheticals, then I don't think we should do it.

carterkozak: Is there an RFC describing the paging design? I'm concerned about the complexity of implementing such a thing in a general way throughout conjure.

iamdanfox: The `Implement batching via paging patterns` as I've seen it done across many services doesn't need any additions to conjure - it's just a convention where people define an endpoint with a `pageToken` optional query param, and then return a `FooPage` which contains an optional `nextPageToken` (the contents of which should remain opaque).

In this specific case, i think the solution would be if the user asked for 3000 ids and the limit for backend requests/responses is 1000, then the first page can just return 1000 and include the pagetoken to say there's more to come.

carterkozak: While easy to misuse, that solution can work for simple batch loads, but it requires users to verify the result otherwise they may fail silently, assuming all results have been read. Unfortunately it doesn't solve batch store operations where the client has sent a full request, and doesn't necessarily expect a response (or expects an atomic operation).

tetigi:
Hello, a delayed update here.

To keep this moving, we'll be stripping back the contentious parts of this RFC and focus instead on the parts that people are in agreement about. In particular, that means that we will move ahead with the following:

- Declaring errors that can be thrown on endpoints, which will improve documentation and improve clarity about what kinds of errors can be thrown where.
- Propagate this error documentation into downstream consumers of Conjure, for example into Javadocs produced by `gradle-conjure`.

As for Java-land new features, we are targeting a reduced set of features that should prove to be useful whilst minimizing contention. More details on those to come in PRs on the relevant repos.

I'll rework the RFC a bit to make it obvious that the scope has been reduced somewhat, and then we can hopefully get this merged.

carterkozak (approved):

PR #835, opened 2021-04-07
Validate conjure headers do not set protocol headers.

jkozlowski:
## Before this PR

Conjure users can set headers that we use for the low level protocol.

## After this PR

==COMMIT_MSG==
Users can no longer set protocol headers (Host, Accept, Content-Type).
==COMMIT_MSG==

## Possible downsides?

Someone is relying on this already?

### DIFF
--- a/conjure-core/src/main/java/com/palantir/conjure/defs/validator/EndpointDefinitionValidator.java
+++ b/conjure-core/src/main/java/com/palantir/conjure/defs/validator/EndpointDefinitionValidator.java
@@ -17,6 +17,7 @@
 package com.palantir.conjure.defs.validator;
 
 import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Sets;
 import com.palantir.conjure.CaseConverter;
 import com.palantir.conjure.either.Either;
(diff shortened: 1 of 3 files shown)


conjure-core/src/main/java/com/palantir/conjure/defs/validator/EndpointDefinitionValidator.java
```diff
@@ -74,6 +75,7 @@ public static void validateAll(EndpointDefinition definition, DealiasingTypeVisi
     }
 
     public static final Pattern HEADER_PATTERN = Pattern.compile("^[A-Z][a-zA-Z0-9]*(-[A-Z][a-zA-Z0-9]*)*$");
+    public static final ImmutableSet<String> PROTOCOL_HEADERS = ImmutableSet.of("Host", "Accept", "Content-Type");
```
jkozlowski, line 78: Maybe tracing headers as well?

carterkozak: I'm less worried about those since they can't break _everything_ heh.

carterkozak (approved):

PR #848, opened 2021-04-21
Fix broken liche checks

tetigi:
## Before this PR
https://www.getpostman.com/ has too large of a header for the `liche` checks. It doesn't look like there's a way to easily expand the read headers (that I can see), so I just combined the postman-collections link into one.

Example of failing build caused by this: https://app.circleci.com/pipelines/github/palantir/conjure/1132/workflows/8965f3da-1ad6-42bc-b4d4-1365648a2247/jobs/9751

## After this PR
==COMMIT_MSG==
Remove docs link to https://www.getpostman.com/ so builds pass again.
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-848.v2.yml
+++ b/changelog/@unreleased/pr-848.v2.yml
@@ -1,0 +1,5 @@
+type: fix
+fix:
+  description: Remove docs link to https://www.getpostman.com/ so builds pass again.
+  links:
+  - https://github.com/palantir/conjure/pull/848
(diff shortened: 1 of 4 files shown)


carterkozak:
I think the link itself is helpful, perhaps we should remove the markdown check? It often causes my builds to fail when third party sites undergo maintenance.

tetigi:
Or we could replace it with another check - the repo in question says that liche is not maintained anymore. Perhaps there are better ones (that also work), I'll take a look.

carterkozak (approved):

changelog/@unreleased/pr-848.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: fix
+fix:
+  description: Replace liche with lychee for markdown link checking
+  links:
+  - https://github.com/palantir/conjure/pull/848
```
carterkozak, line 5: I'd have probably used the `no changelog` tag for this PR since it doesn't impact consumers.

Issue #211, opened 2019-02-01
Cyclic imports break parser


akroy:
## What happened?
a.yml imported an object from b.yml
b.yml imported an object from a.yml
(Note that there's no cycle in the objects. I.e. this would be been fine if it were all in one file)

Parsing crashed. It seems that it recurses, adding `./` until the path gets too long (it failed at 1048 characters, which is suspiciously near 1024, which I'm guessing it a file system limitation)
```
Caused by: shadow.conjure.parser.ConjureParser$ImportNotFoundException: Import not found: /Volumes/git/INTERNAL_PRODUCT/modules/INTERNAL_PRODUCT-api/src/main/conjure/./././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././././a.yml
        at shadow.conjure.parser.ConjureParser$RecursiveParser.parseInternal(ConjureParser.java:101)
        at shadow.conjure.parser.ConjureParser$RecursiveParser.parse(ConjureParser.java:90)
        at shadow.conjure.parser.ConjureParser$RecursiveParser.lambda$parseImports$1(ConjureParser.java:128)
        at shadow.conjure.parser.ConjureParser$RecursiveParser.parseImports(ConjureParser.java:126)
        at shadow.conjure.parser.ConjureParser$RecursiveParser.parseInternal(ConjureParser.java:107)
        at shadow.conjure.parser.ConjureParser$RecursiveParser.parse(ConjureParser.java:90)
        at shadow.conjure.parser.ConjureParser$RecursiveParser.lambda$parseImports$1(ConjureParser.java:128)
        at shadow.conjure.parser.ConjureParser$RecursiveParser.parseImports(ConjureParser.java:126)
        at shadow.conjure.parser.ConjureParser$RecursiveParser.parseInternal(ConjureParser.java:107)
        at shadow.conjure.parser.ConjureParser$RecursiveParser.parse(ConjureParser.java:90)
[...]

PR #948, opened 2021-10-07
Late-resolve imports to enable circular type references

raiju:
## Before this PR

Fixes https://github.com/palantir/conjure/issues/211

Splitting types between files has an impact on parse results. This should have no impact on the generated IR, it only impacts how the conjure can be specified.

## After this PR

==COMMIT_MSG==
Late-resolve imports to enable circular type references
==COMMIT_MSG==


raiju:
This has been broken since at least August 16th (c9d122b1751c09016428ac17ee909f83e4ff26ee)

viktornordling:
I think this actually broke non cyclic imports. Will link more details internally.

PR #956, opened 2021-10-21
Add more examples to conjure definitions docs

domdomegg:
## Before this PR

Not immediately obvious how the default path parameter syntax works.

## After this PR

Adding examples makes the docs more useful.

## Possible downsides?

If I've got this wrong, may be misleading - worth double checking it's correct.


domdomegg:
The CI failure appears to be unrelated to this PR: https://github.com/palantir/conjure/commits/master

xRuiAlves:
Thank you for adding this! 👍

PR #994, opened 2021-11-30
Fix EOF handling and improve error messages of type parsing

raiju:
## Before this PR

Types such as `list<string>>` are accepted (and parsed as `list<string>`; the last char is effectively ignored).

Error messages are really bad:
```
list<string
Expected endToken

at or before character 12
on or before line 0
```

## After this PR

==COMMIT_MSG==
Fix EOF handling and improve error messages of type parsing
==COMMIT_MSG==

EOF in parsing is handled properly.

New error message:
```
Expected end token ">"
list<string
           ^
```


raiju:
Also happy to drop a bunch of unused parsers in the parser sub-package; for now updated them reasonably.

conjure-core/src/main/java/com/palantir/conjure/parser/types/TypeParser.java
```diff
@@ -44,18 +44,24 @@ public ConjureType parse(String input) throws ParseException {
         ParserState inputParserState = new StringParserState(input);
         ConjureType resultType = Parsers.eof(typeParser()).parse(inputParserState);
         if (resultType == null) {
-            throw new ParseException(input, inputParserState);
+            throw new ParseException("Couldn't fully parse input", inputParserState);
         }
-        return parse(new StringParserState(input));
+        return resultType;
```
raiju, line 49: I don't know why everything was parsed twice

conjure-core/src/main/java/com/palantir/parsec/Parsers.java
```diff
@@ -182,11 +201,16 @@ public static ExpectantParser expect(String expectation) {
             public T parse(ParserState input) throws ParseException {
                 T result = parser.parse(input);
 
-                if (input.next() != -1) {
+                if (input.curr() != -1) {
```
raiju, line 204: This fixes the actual parsing bug, everything else is to improve error messages

conjure-core/src/main/java/com/palantir/parsec/StringParserState.java
```diff
@@ -58,12 +62,52 @@ public void release() {
     }
 
     @Override
-    public int getLine() {
-        return 0;
```
raiju, line 62: Joy...

carterkozak (approved):
Nice!

PR #1010, opened 2021-12-09
Late-resolve imports to enable circular type references

raiju:
## Before this PR

https://github.com/palantir/conjure/pull/948 was reverted due to erroneous validation failures. This was caused by paths not being canonicalized properly, causing the same file to be loaded multiple times (e.g. `/.../conjure/id.yml` & `/.../conjure/./id.yml`). This canonicalizes the file in the import & at parsing time.

## After this PR

==COMMIT_MSG==
Late-resolve imports to enable circular type references
==COMMIT_MSG==

## Possible downsides?

- Still testing on repositories to validate there wasn't more than one bug in my approach

### DIFF
--- a/conjure-core/src/main/java/com/palantir/conjure/parser/ConjureParser.java
+++ b/conjure-core/src/main/java/com/palantir/conjure/parser/ConjureParser.java
@@ -23,20 +23,20 @@ import com.fasterxml.jackson.databind.introspect.JacksonAnnotationIntrospector;
 import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
 import com.fasterxml.jackson.datatype.jdk8.Jdk8Module;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableList;
 import com.palantir.conjure.exceptions.ConjureRuntimeException;
 import com.palantir.conjure.parser.types.TypesDefinition;
-import com.palantir.conjure.parser.types.names.Namespace;
-import com.palantir.conjure.parser.types.reference.ConjureImports;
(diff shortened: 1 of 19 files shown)


conjure-core/src/main/java/com/palantir/conjure/parser/ConjureParser.java
```diff
@@ -59,34 +60,71 @@ private ConjureParser() {}
...
+    public static Map<String, AnnotatedConjureSourceFile> parseAnnotated(Collection<File> files) {
+        CachingParser parser = new CachingParser();
+
+        Map<String, AnnotatedConjureSourceFile> parsed = new HashMap<>();
+        Queue<File> toProcess = new ArrayDeque<>(files);
+
+        while (!toProcess.isEmpty()) {
+            File nextFile;
+            try {
+                nextFile = toProcess.poll().getCanonicalFile();
```
raiju, line 86: Fix 1

conjure-core/src/main/java/com/palantir/conjure/parser/types/reference/ConjureImports.java
```diff
@@ -31,21 +36,26 @@
...
-                .file(file)
-                .conjure(conjureDefinition)
-                .build();
+    @Value.Auxiliary
+    @JsonIgnore
+    default ConjureImports resolve(Path baseDir) {
+        try {
+            return ImmutableConjureImports.builder()
+                    .file(file())
+                    .absoluteFile(baseDir.resolve(file()).toFile().getCanonicalFile())
```
raiju, line 55: Fix 2

carterkozak (approved):

PR #1011, opened 2021-12-09
Update docs for Gradle 7

pkoenig10:
## Before this PR
The docs use the removed `compile` Gradle configuration.

## After this PR
The docs use the `implementation` Gradle configuration.


carterkozak (approved):
thanks!

PR #1084, opened 2022-02-25
ImportNotFoundException indicates which file imported the missing file if possible

mcintyret:
Makes debugging large conjure changes much easier!

### DIFF
--- a/conjure-core/src/main/java/com/palantir/conjure/parser/ConjureParser.java
+++ b/conjure-core/src/main/java/com/palantir/conjure/parser/ConjureParser.java
@@ -37,6 +37,7 @@ import java.util.Collection;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Map.Entry;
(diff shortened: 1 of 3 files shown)


conjure-core/src/main/java/com/palantir/conjure/parser/ConjureParser.java
```diff
@@ -79,6 +91,7 @@ public static ConjureSourceFile parse(File file) {
 
         Map<String, AnnotatedConjureSourceFile> parsed = new HashMap<>();
         Queue<File> toProcess = new ArrayDeque<>(files);
+        Map<File, File> importProvenance = new HashMap<>();
```
carterkozak, line 94: Instead of a map, let's create a new simple type for the `toProcess` queue which contains both the `File conjureFile()` and `Optional<File> importedBy()`.

carterkozak:
Resolved in #1088

PR #1099, opened 2022-03-10
Fix references with container prefixes `set`/`list`/`optional`

carterkozak:
Previously importing conjure objects using a reference along the
lines of `settings.Foo` would fail because it would match the
`set` prefix. Now we handle it correctly.

This introduces a change to the way parser markers work:
Instead of maintaining a mark/rewind/release stack, the mark
method returns an object which can be used to return to the
marked location, and the `release` functionality is no longer
necessarily. This takes advantage of the application stack
rather than maintaining a separate stack that can get out of
sync.

==COMMIT_MSG==
Fix references with container prefixes `set`/`list`/`optional`
==COMMIT_MSG==

### DIFF
--- a/conjure-core/src/main/java/com/palantir/parsec/StringParserState.java
+++ b/conjure-core/src/main/java/com/palantir/parsec/StringParserState.java
@@ -16,8 +16,6 @@
 
 package com.palantir.parsec;
 
-import java.util.ArrayDeque;
-import java.util.Deque;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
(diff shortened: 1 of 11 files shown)


conjure-core/src/main/java/com/palantir/parsec/StringParserState.java
```diff
@@ -47,18 +44,9 @@ public int next() {
...
-    public void rewind() {
-        current = marks.pop();
-    }
-
-    @Override
-    public void release() {
-        marks.pop();
+    public MarkedLocation mark() {
+        int markedPosition = current;
+        return () -> current = markedPosition;
```
fawind, line 49: I like this refactoring, makes it easier to reason about the marked locations!

carterkozak: Yep :-) I ran into an issue in my initial fix where a rewind call didn't rewind to the expected mark (resulting in an incorrect value, `map<string` was parsed as `string`), so I figured this refactor would be safer than trying to grok all the mark/rewind pairs!

PR #1140, opened 2022-04-11
Conjure types may be declared with log-safety

carterkozak:
## Before this PR
No real concept of safety aside from hand-waving in ErrorDefinition.

## After this PR
==COMMIT_MSG==
Conjure types may be declared with log-safety. Only conjure primitives (and wrappers around these primitives) may declare safety, complex types calculate safety based on the type graph.
==COMMIT_MSG==

## Possible downsides?
`Map` is not supported at this point, unless the keys and values are wrapped with some other type (Alias).
The current implementation doesn't support declaring safety of external type imports, they're a bit harder to handle due to to lack of information when the IR is compiled.

### DIFF
--- a/conjure-core/src/main/java/com/palantir/conjure/defs/validator/ErrorDefinitionValidator.java
+++ b/conjure-core/src/main/java/com/palantir/conjure/defs/validator/ErrorDefinitionValidator.java
@@ -16,8 +16,11 @@
 
 package com.palantir.conjure.defs.validator;
 
+import com.google.common.collect.ImmutableList;
+import com.palantir.conjure.exceptions.ConjureIllegalStateException;
 import com.palantir.conjure.spec.ErrorDefinition;
 import com.palantir.conjure.spec.FieldDefinition;
+import com.palantir.conjure.spec.FieldName;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -33,5 +36,16 @@ public final class ErrorDefinitionValidator {
                 Stream.concat(definition.getSafeArgs().stream(), definition.getUnsafeArgs().stream())
                         .map(FieldDefinition::getFieldName)
                         .collect(Collectors.toSet()));
+        ImmutableList<FieldName> fieldsDeclaringSafety = Stream.concat(
+                        definition.getSafeArgs().stream(), definition.getUnsafeArgs().stream())
+                .filter(arg -> arg.getSafety().isPresent())
+                .map(FieldDefinition::getFieldName)
+                .collect(ImmutableList.toImmutableList());
+        if (!fieldsDeclaringSafety.isEmpty()) {
+            throw new ConjureIllegalStateException("ErrorDefinition field safety is defined by the key 'safeArgs' or "
+                    + "'unsafeArgs', safety cannot be declared: "
+                    + fieldsDeclaringSafety);
+        }
+        // TODO(ckozak): Resolve safety of all referenced types to ensure alignment with the encapsulating field.
     }
 }
--- a/conjure-core/src/main/java/com/palantir/conjure/defs/validator/ErrorDefinitionValidator.java
+++ b/conjure-core/src/main/java/com/palantir/conjure/defs/validator/ErrorDefinitionValidator.java
@@ -44,8 +44,7 @@ public final class ErrorDefinitionValidator {
         if (!fieldsDeclaringSafety.isEmpty()) {
             throw new ConjureIllegalStateException("ErrorDefinition field safety is defined by the key 'safeArgs' or "
                     + "'unsafeArgs', safety cannot be declared: "
-                    + fieldsDeclaringSafety);
+                    + fieldsDeclaringSafety + " in error " + definition.getErrorName());
         }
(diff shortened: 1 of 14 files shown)


conjure-core/src/main/java/com/palantir/conjure/defs/validator/ErrorDefinitionValidator.java
```diff
@@ -33,5 +36,16 @@ public static void validate(ErrorDefinition definition) {
...
+                        definition.getSafeArgs().stream(), definition.getUnsafeArgs().stream())
+                .filter(arg -> arg.getSafety().isPresent())
+                .map(FieldDefinition::getFieldName)
+                .collect(ImmutableList.toImmutableList());
+        if (!fieldsDeclaringSafety.isEmpty()) {
+            throw new ConjureIllegalStateException("ErrorDefinition field safety is defined by the key 'safeArgs' or "
+                    + "'unsafeArgs', safety cannot be declared: "
+                    + fieldsDeclaringSafety);
+        }
+        // TODO(ckozak): Resolve safety of all referenced types to ensure alignment with the encapsulating field.
```
carterkozak, line 49: Unsure if we want to do this, if so, likely not here. This project doesn't currently need to be aware of how safety is evaluated, only where it's allowed. That would need to change for us to validate at this point -- I can add annotations to the generated errors such that known-unsafe data cannot be passed in, which should be sufficient.

conjure-core/src/main/java/com/palantir/conjure/parser/LogSafetyDefinition.java
```diff
@@ -0,0 +1,52 @@
...
+    DO_NOT_LOG("do-not-log");
+
+    private final String name;
+
+    LogSafetyDefinition(String name) {
+        this.name = name;
+    }
+
+    @JsonCreator(mode = Mode.DELEGATING)
+    public static LogSafetyDefinition parse(String value) {
```
fawind, line 36: Why this extra class instead of directly parsing to `LogSafety`? Is it to support the more human-friendly values (e.g. "do-not-log" instead of "DO_NOT_LOG")?

conjure-core/src/main/java/com/palantir/conjure/defs/validator/ErrorDefinitionValidator.java
```diff
@@ -33,5 +36,15 @@ public static void validate(ErrorDefinition definition) {
                 Stream.concat(definition.getSafeArgs().stream(), definition.getUnsafeArgs().stream())
                         .map(FieldDefinition::getFieldName)
                         .collect(Collectors.toSet()));
+        ImmutableList<FieldName> fieldsDeclaringSafety = Stream.concat(
+                        definition.getSafeArgs().stream(), definition.getUnsafeArgs().stream())
+                .filter(arg -> arg.getSafety().isPresent())
+                .map(FieldDefinition::getFieldName)
+                .collect(ImmutableList.toImmutableList());
+        if (!fieldsDeclaringSafety.isEmpty()) {
+            throw new ConjureIllegalStateException("ErrorDefinition field safety is defined by the key 'safeArgs' or "
```
fawind, line 45: Not sure I get the purpose of this validation. In `ErrorDefinition`'s, we don't allow specifying LogSafety because we already have `safeArgs` and `unsafeArgs` as a first-class concept?

Is it a use-case that people reuse `FieldDefinition` between `ErrorDefinition` and other parts of their API? Afaik, this would limit such reuse? Maybe we want to validate that definitions used as `safeArg` only have safe annotations?

fawind (approved):
Looks good! Just some small questions for my own understanding!

conjure-core/src/main/java/com/palantir/conjure/parser/LogSafetyDefinition.java
carterkozak: That’s correct, we have two apis: the human interaction api (directly parses yaml) and the more structured conjure-defined intermediate representation. This class handles the yaml model where we want to provide consistency with other elements (lower kebab case)

conjure-core/src/main/java/com/palantir/conjure/defs/validator/ErrorDefinitionValidator.java
carterkozak: FieldDefinition from the IR is reused several places within the IR (but does not allow definitions themselves to be used externally). These include : errordefinition safe/unsafe args, object fields, and union types.

I considered breaking out a new type for errors which doesn’t allow safety info, but figured the Abi change wasn’t worth it because we also need to validate that safety isn’t used with complex objects, similar validation for error defs is simple enough. Not a strong opinion, what do you think?

fawind: Thanks for explaining, makes sense. Yeah, think its fine to leave it as-is with validation instead of introducing the new type if this would be a larger change!

carterkozak:
I'm going to hold off merging+releasing this until we've had a chance to review https://github.com/palantir/conjure-java/pull/1753
That way we have more flexibility to update conjure api changes if necessary.

PR #1153, opened 2022-04-19
Implement endpoint argument safety declarations

carterkozak:
## Before this PR
No uniform way to declare safety of endpoint arguments. We use markers in some places (via nasty external type imports), tags in others (`safe` and `unsafe`), and neither of these are documented nor do they align with the type-safe `safety` fields on objects.

## After this PR
==COMMIT_MSG==
Implement consistent endpoint argument safety declarations matching field, alias, and union safety
==COMMIT_MSG==


fawind (approved):
Code looks good and agree with the outcome! Also using `safety` here works quite well and is more intuitive then using markers with external types.

PR #1185, opened 2022-05-16
The parser may enforce safety declarations using `--requireSafety`

carterkozak:
## Before this PR
No way to lock in the safe goodness once an API is fully declared for log-safety

## After this PR
==COMMIT_MSG==
The parser may enforce safety declarations using `--requireSafety`
==COMMIT_MSG==

## Possible downsides?
We may allow safety declarations on external type imports in the future, which may cause projects using this option to fail to upgrade, however we shouldn't gate on perfect before we reach good.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -18,8 +18,8 @@ import net.ltgt.gradle.errorprone.CheckSeverity
 
 buildscript {
     repositories {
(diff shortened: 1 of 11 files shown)


fawind (approved):
Nice!

Issue #1217, opened 2022-07-18
Clarify the serialized wire format for complex error parameters


ash211:
BLUF: we don't specify the wire format for complex error parameters, and conjure-java and conjure-typescript are behaving differently.  We need to determine correct wire format here, and then update the language implementations to match.

-------

Here is the current wire spec for error parameters: https://github.com/palantir/conjure/blob/master/docs/spec/wire.md#55-conjure-errors

> parameters - a JSON map providing additional information regarding the nature of the error.

Unfortunately, the serialization format for these parameters is under-specified, and there's currently a mismatch between what conjure-java generates, and what conjure-typescript expects.  Conjure-java effectively casts complex types to strings and sends the data as a string, whereas conjure-typescript generates .d.ts types expecting the complex types.  Instead, it gets strings.  This manifests as broken typescript code (at runtime!), and a difficulty accessing complex error parameters from typescript.

### Example
Given this yml error definition:

```yml
    errors:
      NoAgentsAvailable:
        namespace: AgentAllocation
        code: CUSTOM_CLIENT
        safe-args:
          unhealthySha256: set<string>
          disabledSha256s: map<string, agentInfoService.AgentDisabledInfo>
          nonExistingSha256s: set<string>
        unsafe-args:
          unhealthy: set<AgentId>
          disabled: map<AgentId, agentInfoService.AgentDisabledInfo>
          nonExisting: set<AgentId>
          message: string
```

The conjure-java library generates errors that look like this (from chrome network tab):

```json
{
    "errorCode": "CUSTOM_CLIENT",
    "errorName": "AgentAllocation:NoAgentsAvailable",
    "errorInstanceId": "682c79a4-b168-4971-acfe-0e58875ad890",
    "parameters": {
        "unhealthySha256": "[]",
        "disabledSha256s": "{}",
        "nonExistingSha256s": "[]",
        "unhealthy": "[]",
        "disabled": "{}",
        "nonExisting": "[]",
        "message": "No healthy and enabled agents available.  Check the status of assigned agents to ensure liveness."
    }
}
```

**Note that the sha256s are strings, and not sets/maps.**

The conjure-generated typescript definition expects these to be sets/maps, and not strings:

```typescript
export interface INoAgentsAvailable {
    'errorCode': "CUSTOM_CLIENT";
    'errorInstanceId': string;
    'errorName': "AgentAllocation:NoAgentsAvailable";
    'parameters': {
        unhealthySha256: Array<string>;
[...]

PR #1218, opened 2022-07-19
Add example for complex error parameters

ash211:
Fixes https://github.com/palantir/conjure/issues/1217

## Before this PR
No example for complex error parameters.  I didn't see it explicitly specified before (see linked issue).

## After this PR
==COMMIT_MSG==
Add example for complex error parameters: a `list<integer>`
==COMMIT_MSG==

## Possible downsides?
Slightly increases size of wire spec document

### DIFF
--- a/docs/spec/wire.md
+++ b/docs/spec/wire.md
@@ -349,6 +349,7 @@ Conjure Errors are serialized as JSON objects with the following keys:
           code: NOT_FOUND
           safe-args:
             name: RecipeName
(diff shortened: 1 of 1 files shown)


pkoenig10:
I'm not convinced this makes the documentation meaningfully clearer. We already have an example for each of the keys listed here and we're not going to add an example for every JSON data type, so it's not clear what is sufficiently special about a list to include it here.

PR #1249, opened 2022-09-19
Report all safety violations

pkoenig10:
## Before this PR
Migrating an existing Conjure definition to add safety declarations is a bit cumbersome because only one error is reported at a time. When going through a Conjure definition, it's easy to miss some of the places where safety declarations are required, especially because not all types/arguments accept safety declarations.

## After this PR
Conjure compilation will collect and report all safety violations instead of throwing an errors as soon as a single violation is discovered. This should make it much easier to users to find all the places that are missing safety declarations.


carterkozak:
Very nice, thank you! :+1:

Issue #1239, opened 2022-09-03
Publish Conjure schema to schemastore.org


gatesn:
## Motivation

Improve IDE support for the human-readable Conjure format.

## Proposal

Publish a JSON schema for the human-readable format to [SchemaStore](https://www.schemastore.org/json/).

IntelliJ, VSCode, and other automatically use these schemas.

carterkozak:
Resolved by #1252

PR #1252, opened 2022-09-20
Create Conjure JSON schema

gatesn:
## Before this PR
No IDE support at all for writing Conjure human-readable format

## After this PR
JSON schema exists for the human readable format. I will follow up with a PR to schemastore.org to add this to their catalog.

See #1239

We get some auto-completion in IntelliJ:
image

And some basic validation checks:
image

## Possible downsides?
This format is defined only in Jackson/Immutables at the moment, so it can't easily be automatically converted to JSON without losing information. It's therefore (painstakingly!) hand-crafted. I figure it's a slow-moving spec, so shouldn't be an issue given the benefits we get from this.

### DIFF
--- a/conjure.schema.json
+++ b/conjure.schema.json
@@ -1,0 +1,525 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "type": "object",
+  "additionalProperties": false,
+  "properties": {
+    "types": {
+      "$ref": "#/definitions/TypesDefinition"
+    },
+    "services": {
+      "type": "object",
+      "additionalProperties": {
+        "$ref": "#/definitions/ServiceDefinition"
+      }
+    }
+  },
+  "definitions": {
+    "TypesDefinition": {
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "conjure-imports": {
+          "type": "object",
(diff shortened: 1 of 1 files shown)


robert3005:
👍

changelog/@unreleased/pr-1252.v2.yml
```diff
@@ -0,0 +1,24 @@
...
+
+    We get some auto-completion in IntelliJ:
+    <img width="696" alt="image" src="https://user-images.githubusercontent.com/1928052/191144274-330e7785-ffd2-4184-95e9-28e5358ff88b.png">
+
+    And some basic validation checks:
+    <img width="654" alt="image" src="https://user-images.githubusercontent.com/1928052/191144243-2bc3f6f0-d454-4199-bcee-267b5f598f34.png">
+
+
+    ## Possible downsides?
+    This format is defined only in Jackson/Immutables at the moment, so it can't easily be automatically converted to JSON without losing information. It's therefore (painstakingly!) hand-crafted. I figure it's a slow-moving spec, so shouldn't be an issue given the benefits we get from this.
```
carterkozak, line 22:
```suggestion
    Define JSON schema for the Conjure yaml format
```

carterkozak (approved):

PR #1266, opened 2022-09-30
adding support for log safety annotations on certain external imports 

lsingh123:
Adds support for log safety annotations on certain kinds of external imports (specifically, Java boxed primitives) at import time. The goal is to (once we make a parallel change in the conjure generator) enable code snippets like the following:

```
  imports:
    RealmId:
      base-type: string
      external:
        java: java.lang.Long
      safety: safe
    SensitiveLong:
      base-type: string
      external:
        java: java.lang.Long
      safety: unsafe
```

To make rollout and enforcement as smooth as possible, this PR adds a configuration option to opt-in to this feature. For discussion of this config, see [this](https://github.com/palantir/conjure/pull/1266#discussion_r1084654768) comment.


conjure-core/src/test/java/com/palantir/conjure/defs/validator/ConjureSourceFileValidatorTest.java
```diff
@@ -362,88 +369,6 @@ public void testInvalidSafetyArgument_reference() {
                                 + " conjure primitives may declare safety. package.Foo is not a primitive type.");
     }
 
-    @Test
-    public void testValidArgumentSafety() {
```
lsingh123, line 366: these cases are covered more comprehensively by the parameterized tests below.

conjure-core/src/test/java/com/palantir/conjure/defs/validator/ConjureSourceFileValidatorTest.java
```diff
@@ -510,11 +437,297 @@ public void testSafetyMarkersAreNotAllowedWhenSafetyIsRequired() {
...
     private FieldDefinition field(FieldName name, String type) {
         return FieldDefinition.builder()
                 .fieldName(name)
                 .type(Type.reference(TypeName.of(type, PACKAGE)))
                 .docs(DOCS)
                 .build();
     }
+
+    private static Stream<Arguments> getAllTypesToTest_SafetyAtImportTime(Type externalReference) {
+        ConjureDefinition conjureDefObject = ConjureDefinition.builder()
```
lsingh123, line 617: assumes externalReference has a safety annotation

conjure/src/main/java/com/palantir/conjure/cli/ConjureCli.java
```diff
@@ -185,6 +189,19 @@ static void generate(CliConfiguration config) {
...
 
+        private static SafetyDeclarationRequirements getSafetyRequirements(CliConfiguration config) {
+            if (config.requireSafety() && config.requireSafetyExternalImports()) {
+                return SafetyDeclarationRequirements.EXTERNAL_IMPORTS_REQUIRED;
+            } else if (config.requireSafety() && !config.requireSafetyExternalImports()) {
+                return SafetyDeclarationRequirements.REQUIRED;
+            } else if (!config.requireSafety() && !config.requireSafetyExternalImports()) {
+                return SafetyDeclarationRequirements.ALLOWED;
+            } else {
+                System.err.println("If safety is required for external types, it must be required for all types.");
```
lsingh123, line 200: Because I'm using two booleans to represent three potential cases (safety allowed, safety required, safety required for external types), there's this awkward 4th case that isn't quite correct. We could avoid this weird fourth case by instead migrating the config to an enum with 3 options. 

I avoided the enum because it would take a lot of upfront work to migrate all the existing conjure configurations to an enum, to say nothing of non-palantir usages that we couldn't excavate. On the other hand, the enum gives us more flexibility, which may be important if we want similar configurability as we expand this feature to cover more kinds of external imports in the future.

An in-between solution is to deserialize the requireSafety() config options into a String instead of a boolean, and then map as following:
true -> REQUIRED
false -> ALLOWED
EXTERNAL_IMPORTS_REQUIRED -> EXTERNAL_IMPORTS_REQUIRED
REQUIRED -> REQUIRED
ALLOWED -> ALLOWED

This would give us future config flexibility and is backwards compatible, but is more counterintuitive than the two above options to me. @carterkozak thoughts?

conjure-core/src/main/java/com/palantir/conjure/defs/validator/SafetyValidator.java
```diff
@@ -183,6 +217,18 @@ public Stream<String> visitReference(TypeName value) {
...
         public Stream<String> visitExternal(ExternalReference value) {
+            String importType = getImportType(value);
+            if (ValidImportsForSafety.ALLOWED_IMPORTS.containsKey(importType)) {
+                Type givenFallbackType =
+                        Type.primitive(PrimitiveType.valueOf(ValidImportsForSafety.ALLOWED_IMPORTS.get(importType)));
+                if (givenFallbackType.equals(value.getFallback())) {
+                    return Stream.empty();
+                } else {
+                    return Stream.of(String.format(
+                            "Mismatched base type. %s must have a base type of %s in order to declare safety.",
```
lsingh123, line 228: I'm not convinced that this is the best place for this check - as a user, I might be confused if I suddenly start getting a mismatched fallback type error when adding safety annotations. We could just skip this check, unless there's particular harm in allowing safety annotations for external imports with an invalid fallback type.

lsingh123: @nathklei you suggested this check - do you think we absolutely need it?

conjure-core/src/main/java/com/palantir/conjure/defs/validator/ValidImportsForSafety.java
```diff
@@ -0,0 +1,37 @@
...
+
+import java.util.Map;
+
+final class ValidImportsForSafety {
+
+    /*
+    Limiting to boxed primitives for now, but would eventually want to support more external imports.
+     */
+    static final Map<String, String> ALLOWED_IMPORTS = Map.ofEntries(
+            Map.entry("java.lang.Long", "string"),
```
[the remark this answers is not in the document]
lsingh123: sounds good to me. was thinking that this might help address @carterkozak's concern that _just_ supporting Longs might confused users, but maybe this is a roundabout solution to that problem.

conjure-core/src/main/java/com/palantir/conjure/defs/SafetyDeclarationRequirements.java
```diff
@@ -18,9 +18,23 @@
...
+        return this == EXTERNAL_IMPORTS_REQUIRED;
+    }
+
+    public boolean allowed() {
+        return this == ALLOWED;
+    }
+
+    public boolean strictMode() {
+        return this == REQUIRED || this == EXTERNAL_IMPORTS_REQUIRED;
+    }
```
carterkozak, line 39: Thoughts on adding support for safety in a subset of external type imports without changing the validation logic (not in the same commit, anyhow) to keep the change small? If memory serves, we have a custom validation module internally for the project which will benefit most from this, which already limits the allowed external type imports -- it may be better suited to enforce that all valid imports declare safety.

conjure-core/src/main/java/com/palantir/conjure/defs/validator/ValidImportsForSafety.java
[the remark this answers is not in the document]
carterkozak: I recall BigInteger uses as well

conjure-core/src/main/java/com/palantir/conjure/defs/validator/ValidImportsForSafety.java
```diff
@@ -0,0 +1,29 @@
...
+package com.palantir.conjure.defs.validator;
+
+import java.util.List;
+
+final class ValidImportsForSafety {
+
+    /*
+    Limiting to java.lang.Long for now, but would eventually want to support more external imports.
+     */
+    static final List<String> ALLOWED_IMPORTS = List.of("java.lang.Long");
```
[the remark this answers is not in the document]
lsingh123: It looks like there are quite a few instances of people importing things from `com.palantir` packages: https://sourcegraph.palantir.build/search?q=context:global+external:+file:.yml+-repo:%5Egithub%5C.com/palantir/conjure&patternType=standard&sm=1&groupBy=repo Ignoring `com.palantir.logsafe.Safe` (which I think is the old way of marking safety?), I see lots of random types that might be useful to log like ResourceIdentifier, ObjectRid, User, GeotimeTrackGid, CosmosPlanGid, etc. I was imagining longs/primitives as a first step, and then later supporting safety annotations on most external imports, so static list for now, more complicated logic later.

conjure-core/src/main/java/com/palantir/conjure/defs/SafetyDeclarationRequirements.java
lsingh123: That makes sense to me. Would we do the validation here later, or only perform validation in the internal module? Also, could you point me to that internal module?

carterkozak: Sent internally :-)

> Would we do the validation here later, or only perform validation in the internal module?

I think this depends on rollout, and whether we intend to support additional types of external type imports later on. For the foreseeable future, it's probably best if we assume only the internal module will handle validation that all external type imports of `java.lang.Long` declare safety.

conjure-core/src/main/java/com/palantir/conjure/defs/validator/ValidImportsForSafety.java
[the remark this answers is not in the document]
carterkozak: 👍 That sounds good to me.

`ResourceIdentifier` is a funny one because conjure provides a first-class `rid` type using the same class! The other types are examples that we probably don't want to allow safety declarations on because the types themselves already declare their own safety -- it wouldn't be ideal if the external import declared a type as `safe` when the java class is annotated `@Unsafe`.

lsingh123:
going to make the parsing-only changes in a separate branch

PR #1268, opened 2022-10-03
Fix Conjure enum mixing short-hand and fully-specified value types

gatesn:
## Before this PR

For an array of `oneOf` types, JSON Schema picks a single type that the whole array must match. I'm not sure if this is correct JSON schema or just the implementation we're using. It seems a little odd?

In any case, the example fails the json schema test with:
```
Unexpected JSON schema error 'oneOf' $.types.definitions.objects.MyEnum.values[1]: should be valid to one and only one of schema, but more than one are valid: {"$ref":"#/definitions/UpperCase"}{"$ref":"#/definitions/EnumValueDefinition"}
```

## After this PR

The example passes the test.

I've also renamed the directory the example YML files are in so they're picked up as Conjure files by the IDE.


carterkozak:
Please fix the changelog

carterkozak (approved):
Thanks!

PR #1273, opened 2022-10-04
UpperCase should allow numbers

j-baker:
Right now this complains on e.g. ACCUMULATOR_V2 because it doesn't like the 2. The replacement is the actual validation that Conjure does.


j-baker:
image

fawind (approved):

PR #1315, opened 2023-01-31
support parsing the safety of external imports

lsingh123:
as discussed with @carterkozak, this PR supports adding safety annotations for external imports without performing any validation logic


conjure-core/src/main/java/com/palantir/conjure/defs/ConjureParserUtils.java
```diff
@@ -393,7 +403,7 @@ static List<FieldDefinition> parseField(
                 .collect(Collectors.toList());
     }
 
-    private static LogSafety parseLogSafety(LogSafetyDefinition def) {
+    public static LogSafety parseLogSafety(LogSafetyDefinition def) {
```
carterkozak, line 406: Are there any uses of this method outside this class? We may be able to keep it private

lsingh123: It's used in ConjureTypeParserVisitor, but I can make it package-private

lsingh123: oh yes, you're totally right 🤦 fixing

conjure-core/src/test/resources/example-external-types.yml
```diff
@@ -8,6 +8,11 @@ types:
       base-type: string
       external:
         java: com.palantir.example.ExternalIntegerImport
+    ExampleLong:
+      base-type: string
+      external:
+        java: java.lang.Long
```
carterkozak, line 14: To be super safe, we may want to add validation that using `base-type: bearertoken` cannot be marked safe/unsafe because it's do-not-log.

lsingh123: added some validation logic for that!

conjure-core/src/main/java/com/palantir/conjure/defs/ConjureParserUtils.java
```diff
@@ -191,6 +193,25 @@ public static Type parsePrimitiveType(com.palantir.conjure.parser.types.primitiv
         return Type.primitive(PrimitiveType.valueOf(primitiveType.name()));
     }
 
+    static Type parseExternalType(ExternalTypeDefinition externalType, String conjurePackage, String typeName) {
+        if (isBearerToken(externalType)) {
```
carterkozak, line 197: I think this will fail if any external types declare base-type bearer-token
```suggestion
        if (isBearerToken(externalType) && externalType.safety().isPresent()) {
```

lsingh123: oops, thanks for the catch

carterkozak (approved):

PR #1318, opened 2023-02-01
don't fail on external imports

lsingh123:
validation for external imports is performed in internal code, but shouldn't automatically fail in the parser

### DIFF
--- a/conjure-core/src/test/java/com/palantir/conjure/defs/validator/ConjureSourceFileValidatorTest.java
+++ b/conjure-core/src/test/java/com/palantir/conjure/defs/validator/ConjureSourceFileValidatorTest.java
@@ -511,6 +511,28 @@ public class ConjureSourceFileValidatorTest {
                 .hasMessageContaining("Service.end(arg): Safety markers have been replaced by the 'safety' field");
     }
 
+    @Test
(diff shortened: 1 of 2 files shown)


conjure-core/src/test/java/com/palantir/conjure/defs/validator/ConjureSourceFileValidatorTest.java
```diff
@@ -511,6 +511,28 @@ public void testSafetyMarkersAreNotAllowedWhenSafetyIsRequired() {
...
+                        .typeName(FOO)
+                        .fields(FieldDefinition.builder()
+                                .fieldName(FieldName.of("bad"))
+                                .type(Type.external(external))
+                                .safety(LogSafety.UNSAFE)
+                                .docs(DOCS)
+                                .build())
+                        .build()))
+                .build();
+        ConjureDefinitionValidator.validateAll(conjureDef, SafetyDeclarationRequirements.REQUIRED);
```
carterkozak, line 533: Isn’t this unsafe? The external type is declared do-not-log, but it’s later declared as unsafe instead when a field is defined.

I don’t think we want to allow type safety from external type imports to be overridden when the type is used in a field.

PR #1319, opened 2023-02-02
add a more descriptive error message for external import failures

lsingh123:
and some tests confirming expected behavior

### DIFF
--- a/conjure-core/src/main/java/com/palantir/conjure/defs/validator/SafetyValidator.java
+++ b/conjure-core/src/main/java/com/palantir/conjure/defs/validator/SafetyValidator.java
@@ -186,7 +186,9 @@ public final class SafetyValidator {
             String referenceName = value.getExternalReference().getPackage() + "."
                     + value.getExternalReference().getName();
             return Stream.of(String.format(
-                    "External import %s must declare safety at import time, not at usage time.", referenceName));
+                    "If external import %s is eligible to declare safety, it must do so at import time, not at usage"
+                            + " time.",
+                    referenceName));
         }
 
         @Override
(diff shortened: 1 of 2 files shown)


conjure-core/src/main/java/com/palantir/conjure/defs/validator/SafetyValidator.java
```diff
@@ -183,7 +183,10 @@ public Stream<String> visitReference(TypeName value) {
 
         @Override
         public Stream<String> visitExternal(ExternalReference value) {
-            return Stream.of(fail(parentReference, value.getExternalReference()));
+            String referenceName = value.getExternalReference().getPackage() + "."
+                    + value.getExternalReference().getName();
+            return Stream.of(String.format(
+                    "External import %s must declare safety at import time, not at usage time.", referenceName));
```
carterkozak, line 189: hmm, this creates the impression that any external type reference can declare safety. Wording is hard.

lsingh123: I think for now the only ineligible external imports are bearer tokens (at least in the parser, ignoring the internal validation module), which fail parsing because of this [change](https://github.com/palantir/conjure/pull/1315/files#diff-55c3a35aaa4e3a489625722cdb17a78ed8f6d9221d6c0a31d6114522a3a2ac23). So they'll only hit this case if it's an eligible external type? I could say something like "If external import %s is eligible to declare safety, it must do so at import time not at usage time."

carterkozak: Ah you're right, I'd thought we still had the java.lang.Long restriction here, but that's not the case.

carterkozak (approved):

PR #1386, opened 2023-05-09
Improve error message for failed case conversion

ash211:
## Before this PR
When attempting to build graphql schema with a field element of `egressCIDRs` this error message appears:
```
Unexpected case for: egressCIDRs
```

I was confused what this meant, and thought it was related to an unhandled switch-case entry.  Instead, it's actually a problem that `egressCIDRs` isn't in a standard casing for conjure, and I fixed by renaming the graphql field to `egressCidrs`.

## After this PR
==COMMIT_MSG==
Improve error message for failed case conversion
==COMMIT_MSG==

The new message is:
```
Unexpected case for: 'myCIDRs'. Expected one to be in one of these case formats: [LOWER_CAMEL_CASE, KEBAB_CASE, SNAKE_CASE]
```

When seeing this, I think I would've been clued in to changing `egressCIDRs` -> `egressCidrs`

## Possible downsides?
Throwing this exception is slightly more expensive now, but this ought not be in any hot paths.

### DIFF
--- a/conjure-generator-common/src/main/java/com/palantir/conjure/CaseConverter.java
+++ b/conjure-generator-common/src/main/java/com/palantir/conjure/CaseConverter.java
@@ -18,7 +18,6 @@ package com.palantir.conjure;
 
 import com.google.common.base.CaseFormat;
 import com.google.errorprone.annotations.Immutable;
-import java.util.List;
 
 @Immutable
 public final class CaseConverter {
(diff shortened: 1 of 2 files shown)


conjure-generator-common/src/main/java/com/palantir/conjure/CaseConverter.java
```diff
@@ -99,6 +101,8 @@ private static Case nameCase(String name) {
                 return nameCase;
             }
         }
-        throw new IllegalArgumentException("Unexpected case for: " + name);
+        throw new IllegalArgumentException("Unexpected case for: '" + name
+                + "'. Expected one to be in one of these case formats: "
+                + Arrays.stream(Case.values()).map(Enum::name).collect(Collectors.toList()));
```
carterkozak, line 106: Could we convert each element from screaming snake into itself for additional inline docs? e.g. `[lowerCamelCase, kebab-case, snake_case]`

carterkozak: We could also hard code the string, either way works for me :-]

ash211: I had originally wanted to avoid that so the message got automatically updated for any new entries to the enum, but lost that when going to Guava.  So we might as well go all the way and just hardcode the string!

carterkozak (approved):

PR #1404, opened 2023-06-13
[Improvement] Add validation for supported HTTP methods

nickbar01234:
## Before this PR

The compiler accepts unspecified HTTP methods when generating the IR. For example, the input Conjure definition

```yml
...
    endpoints:
      get:
        # Method BOGUS does not exist
        http: BOGUS /my-service
```

produces

```yml
...
    "endpoints" : [ {
      "endpointName" : "get",
      "httpMethod" : "BOGUS",
      "httpPath" : "/my-service",
      "args" : [ ],
      "markers" : [ ],
      "tags" : [ ]
    } ]
```

## After this PR

An additional validator is added in the semantic-check when parsing `EndpointDefinition`.

==COMMIT_MSG==
==COMMIT_MSG==

### DIFF
--- a/conjure-core/src/main/java/com/palantir/conjure/defs/validator/EndpointDefinitionValidator.java
+++ b/conjure-core/src/main/java/com/palantir/conjure/defs/validator/EndpointDefinitionValidator.java
@@ -27,7 +27,6 @@ import com.palantir.conjure.spec.ArgumentName;
 import com.palantir.conjure.spec.EndpointDefinition;
 import com.palantir.conjure.spec.ExternalReference;
 import com.palantir.conjure.spec.HttpMethod;
-import com.palantir.conjure.spec.HttpMethod.Value;
 import com.palantir.conjure.spec.ListType;
 import com.palantir.conjure.spec.MapType;
 import com.palantir.conjure.spec.OptionalType;


conjure-core/src/main/java/com/palantir/conjure/defs/validator/EndpointDefinitionValidator.java
```diff
@@ -457,6 +458,19 @@ public void validate(EndpointDefinition definition) {
         }
     }
 
+    @com.google.errorprone.annotations.Immutable
+    private static final class NoUnsupportedHttpMethodValidator implements ConjureValidator<EndpointDefinition> {
+        @Override
+        public void validate(EndpointDefinition definition) {
+            Preconditions.checkState(
+                    HttpMethod.values().contains(definition.getHttpMethod()),
```
carterkozak, line 466: I think we can simplify this into `!HttpMethod.Value.UNKNOWN.equals(definition.getHttpMethod().value())`. Our generated enums wrap standard java enums, where we use an additional `UNKNOWN` value to represent any string value that isn't declared in the enum.

nickbar01234: Got it, thanks for the suggestion 😄.

carterkozak (approved):
Thanks!

Issue #1447, opened 2023-09-10
Validator failing to validate nested type definitions


jj-marinho:
## What happened?
Some conjure definitions that look like they shouldn't be allowed according to validation rules, are being allowed if they happen nested within a type.

The 2 main errors seem to be:
1. `map<ValidKey, AnyInvalidDefinition>` such as `map<ValidKey, optional<optional<map<InvalidKey, string>>>>`. Here 2 nested optionals should trigger a invalid compilation and a map with an invalid key also should trigger.
2. type: `optional<InvalidDefinition>` (besides the case `optional<optional<...>>` that has it's own validator), such as `optional<map<InvalidKey, string>>`.

And I suspect there might be more.

Simply creating a conjure definition containing the former types is enough to cause the compilations mistakes.

The errors are related to how ConjureDefinitionValidator works. Where it fails to recursively check definitions on many cases.

### Examples
Looks like map value error could simply be a miswritten code line, as follows: https://github.com/palantir/conjure/pull/1448. Even then, creating a `map<string, optional<optional<...>>> might still be allowed, even though it shouldn't.

The type inside an optional error might require a larger refactor: https://github.com/palantir/conjure/pull/1449

## What did you want to happen?
Both cases outlined above should throw when validationg.

ConjureDefinitionValidator should be able to recursively apply validations for types within types. Unwrapping and recursing when necessary.

PR #1448, opened 2023-09-10
Fixing map key validation

jj-marinho:
## Before this PR
One of the examples of: https://github.com/palantir/conjure/issues/1447

Map key validation didn't recursively look inside map value types correctly.

## After this PR
Now it does check inside the map value.

==COMMIT_MSG==
Fixing map key validation
==COMMIT_MSG==


carterkozak (approved):
thanks!

PR #1516, opened 2024-02-29
Add inProcessExecution that propagates the conjure exceptions

crogoz:
## Before this PR

Conjure exceptions were not propagated to gradle-conjure. We were only getting the `exitCode`. 
- we were converting any exception into an exitCode: https://github.com/palantir/conjure/blob/f2cf6baa77d78b0aa2e6f3c1343542904756c1ee/conjure/src/main/java/com/palantir/conjure/cli/ConjureCli.java#L58C45-L58C46
- in gradle-conjure we were only getting the exitCode: https://github.com/palantir/gradle-conjure/blob/003f0e5b4c2d928a49183c762ec4f09ed14573d2/gradle-conjure/src/main/java/com/palantir/gradle/conjure/ConjureRunnerResource.java#L198 

before the exception looked like:
```
* What went wrong:
Execution failed for task ':types-api:rawIr'.
> Failed to generate conjure IR. The command 'conjure' with args [compile, /Volumes/git/service/types-api/build/conjure, /Volumes/git/service/types-api/build/conjure-ir/rawIr.conjure.json, --extensions, {"recommended-product-dependencies":[]}, --requireSafety] failed with exit code -1. Output above.
```

## After this PR
Introduces a new `inProcessExecution ` method that will re-throw any exception

```
* What went wrong:
Execution failed for task ':types-api:rawIr'.
> com.palantir.conjure.exceptions.ConjureIllegalStateException: Unknown LocalReferenceType: TypeName{name=Name1}
  
  Failed to generate conjure IR. The command '[compile, /Volumes/git/service/types-api/build/conjure, /Volumes/git/service/types-api/build/conjure-ir/rawIr.conjure.json, --extensions, {"recommended-product-dependencies":[]}, --requireSafety]' failed.
```
==COMMIT_MSG==
Add inProcessExecution that propagates the conjure exceptions
==COMMIT_MSG==


conjure/src/main/java/com/palantir/conjure/cli/ConjureCli.java
```diff
@@ -58,61 +57,81 @@ public static void main(String[] args) {
...
-                } else {
-                    commandLine.getErr().println(exception.getMessage());
-                }
-            });
+            String message = chain.stream()
+                    .map(exception -> (exception instanceof JsonMappingException)
+                            ? handleJsonMappingException((JsonMappingException) exception)
+                            : exception.getMessage())
+                    .collect(Collectors.joining("\n"));
+            commandLine.getErr().println(message);
```
CRogers, line 110: What happens to this when called in process? Does it still print out to stderr?

CRogers: Looks like there is a `setErr` we can use on `CommandLine` to change it.

crogoz: Screenshot 2024-03-04 at 2 54 22 PM
Yes, it does still log to stderr. the getErr() is the default standardError: https://github.com/remkop/picocli/blob/main/src/main/java/picocli/CommandLine.java#L1311C16-L1311C19

conjure/src/main/java/com/palantir/conjure/cli/ConjureCli.java
```diff
@@ -58,61 +57,80 @@ public static void main(String[] args) {
...
+    }
+
     @VisibleForTesting
     static CommandLine prepareCommand() {
         return new CommandLine(new ConjureCli()).setExecutionExceptionHandler(new ExceptionHandler());
     }
 
     public static final class ExceptionHandler implements IExecutionExceptionHandler {
+
+        private Optional<Exception> thrownException = Optional.empty();
```
CRogers, line 75: I actually think we can write this a little more concisely/safely for future changes. We can keep the existing `IExecutionExceptionHandler` as is then write a wrapper `IExecutionExceptionHandler` that wraps the original one but catches the exception and saves it. Also sets the `err` and if that is written to, we make an exception from that.

CRogers: Perhaps with a `TeeOutputStream` to the original `err` if we want to preserve the current behaviour (maybe we don't need to though?)

crogoz: added a separate class that does the exception wrapping here: https://github.com/palantir/conjure/pull/1516/files#diff-cc1e8eaeed51e87c5b283c659588f4c2f976ceafdc8ad587e89fc92460fdccedR33.

CRogers: Nice!

CRogers (dismissed):

CRogers (approved):

PR #1563, opened 2024-03-25
Setup NPM Token using generateNpmRc Task

bmarcaur:
## Before this PR
We assumed the npmrc location and contents, but they changed long ago.

## After this PR
`generateNpmRc` has configuration points to provide a token directly.

### DIFF
--- a/conjure-api/build.gradle
+++ b/conjure-api/build.gradle
@@ -26,9 +26,7 @@ project (':conjure-api:conjure-api-objects') {
 }
 
 project (':conjure-api:conjure-api-typescript') {
-    publishTypeScript.doFirst {
-        file('src/.npmrc') << "//registry.npmjs.org/:_authToken=${System.env.NPM_AUTH_TOKEN}"
-    }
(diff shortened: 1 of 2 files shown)


bjlaub (dismissed):

conjure-api/build.gradle
```diff
@@ -26,9 +26,7 @@ project (':conjure-api:conjure-api-objects') {
 }
 
 project (':conjure-api:conjure-api-typescript') {
-    publishTypeScript.doFirst {
-        file('src/.npmrc') << "//registry.npmjs.org/:_authToken=${System.env.NPM_AUTH_TOKEN}"
-    }
+    generateNpmrc.token = System.env.NPM_AUTH_TOKEN;
```
carterkozak, line 29: Do we need to update this as well? https://github.com/palantir/conjure/blob/731f08dacd665410b6fb6ed00fd29ae477d63fc0/docs/howto/publish_typescript_to_npm.md

bmarcaur: Yes we do

bjlaub (approved):

PR #1570, opened 2024-04-08
ServiceDefinition name field may be omitted

carterkozak:
## Before this PR
`ServiceDefinition.name` has been deprecated for 6 years.

## After this PR
==COMMIT_MSG==
ServiceDefinition name field may be omitted
==COMMIT_MSG==

### DIFF
--- a/conjure.schema.json
+++ b/conjure.schema.json
@@ -390,7 +390,6 @@
         }
       },
       "required": [
-        "name",
         "package",
         "endpoints"
       ]
--- a/changelog/@unreleased/pr-1570.v2.yml
+++ b/changelog/@unreleased/pr-1570.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: ServiceDefinition name field may be omitted
+  links:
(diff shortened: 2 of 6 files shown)


ash211 (dismissed):
Thank you!

Please also remove the REQUIRED tag at https://github.com/palantir/conjure/blob/061f39778fc3b63117411b0042c5971e9352ae5f/docs/spec/conjure_definitions.md#L303

carterkozak:
Good call, I've removed the reference to the `name` field entirely because the value is completely unused

conjure.schema.json
```diff
@@ -390,7 +390,6 @@
         }
       },
       "required": [
-        "name",
```
ash211, line 393: Seems like circle was green even prior to removing `name` from this required list.  Should we add a test that all of the valid test conjure yml files in https://github.com/palantir/conjure/tree/master/conjure-core/src/test/resources successfully validate against this conjure schema?  Fine to do as a FLUP, IMO

pkoenig10 (approved):

PR #1615, opened 2024-08-07
[HackWeek] Add default values for (select) primitives and enums to the IR compiler

cflems:
## Before this PR
Services using Conjure that wish to remain backwards-compatible while expanding their storage types must declare all new types as `Optional`s in order to avoid breaking Conjure's deserialization. This requires defaults to be declared in the implementation layer rather than in the API layer, in addition to creating a large number of useless box objects that bloat memory, and confusing the API spec with optional objects that are not actually optional. I personally have encountered bugs where a value was accidentally omitted from a "for backwards-compatibility only" in a storage type, causing developer hours to be wasted on something that should have been caught by the type system.

## After this PR
[FieldDefinition](https://palantir.github.io/conjure/#/docs/spec/conjure_definitions?id=fielddefinition)s in YAML now support a `default` parameter which accepts a value to supplied when null or empty values of this field are deserialized instead of throwing an error. Currently all primitive types (barring bearertokens and `ANY`) and enum values are supported. More complex types would work too, but the demand is small and I want to constrain the scope of this project.

==COMMIT_MSG==
Add default values for primitives and enums to the IR compiler
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-1615.v2.yml
+++ b/changelog/@unreleased/pr-1615.v2.yml
@@ -1,0 +1,5 @@
+type: feature
+feature:
+  description: Add default values for primitives and enums to the IR compiler
+  links:
+  - https://github.com/palantir/conjure/pull/1615
--- a/conjure-api/src/main/conjure/conjure-api.yml
+++ b/conjure-api/src/main/conjure/conjure-api.yml
@@ -26,6 +26,17 @@ types:
       Documentation:
         alias: string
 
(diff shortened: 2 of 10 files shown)


cflems:
Enum aliases may need to die on second thought. I don't think we want the code gen to do `Type.of(Type.of(Type.of(Type.of(Type.of(Type.of())))))` and I also don't think anyone wants it.

PR #1661, opened 2024-10-17
[Feature] Allow ConjureTypes to be the 'base-type' on an External Import

bmarcaur:
## Before this PR
According to the [definition spec for `ExternalTypes`](https://github.com/palantir/conjure/blob/master/docs/spec/conjure_definitions.md#externaltypedefinition), it is a valid definition to allow for any [`ConjureType` which includes `ContainerType`](https://github.com/palantir/conjure/blob/master/docs/spec/conjure_definitions.md#conjuretype). However, when I wrote the following definition:

```yaml
types:
  imports:
    Flags:
      base-type: list<string>
      external:
        java: com.palantir.Flags
```

It errored out with the following:

```
Cannot construct instance of `com.palantir.conjure.parser.types.primitive.PrimitiveType`, problem: TypeNames must be a primitive type [datetime, boolean, string, double, bearertoken, binary, safelong, integer, rid, any, uuid] or match pattern ^[A-Z][a-z0-9]+([A-Z][a-z0-9]+)*$: list<string>
  @ types -> imports -> Flags -> base-type
TypeNames must be a primitive type [datetime, boolean, string, double, bearertoken, binary, safelong, integer, rid, any, uuid] or match pattern ^[A-Z][a-z0-9]+([A-Z][a-z0-9]+)*$: list<string>
```

## After this PR
Changes the `base-type` deserialization object to allow for all valid `ConjureTypes`. This addresses a concern where you want to optimize the serialization of a type in Java, but you are also adhering to the wire spec and thus have a valid conjure compliant serialization i.e. it is safe to convey that type to other generators instead of defaulting to `any`.

## Possible downsides?
Spec is one thing, observed behavior is another. I think because this is a strict expansion of the optionality this _might_ be safe. As in it should not break anyone because every existing instance is a `PrimitiveType` which is a subset of `ConjureType`. However, what this now means is that the following complex setup is valid:

```yaml
types:
  imports:
    Flags:
      base-type: list< ComplexFlags >
      external:
        java: com.palantir.Flags
        
  definitions:
    default-package: com.palantir
    objects:
      ComplexFlags:
        fields:
          field1: string
          field2: string
```

I'm not sure if this is an issue. I need to consider it more. My initial thought is that this is fine, because the intent of `base-type` is that it be a type that every language can represent. Which is tautologically true given the nature of Conjure. And `ConjureType` is strictly either primitive, collection, or typename and specifically cannot reference another `ExternalType`.

I also did some quick searching for usages and there is only one place that relies on the return type of `baseType()` so this should not be that disruptive from that perspective.

## ToDo

- [ ] Compare the generated IR with a locally published version
- [ ] Consider what consumers of this IR may or may not be expecting

### DIFF
--- a/changelog/@unreleased/pr-1661.v2.yml
+++ b/changelog/@unreleased/pr-1661.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: '[FR] Allow ConjureTypes to be the ''base-type'' on an External Import'
+  links:
+  - https://github.com/palantir/conjure/pull/1661
--- a/conjure-core/src/main/java/com/palantir/conjure/defs/ConjureParserUtils.java
+++ b/conjure-core/src/main/java/com/palantir/conjure/defs/ConjureParserUtils.java
@@ -193,15 +193,20 @@ public final class ConjureParserUtils {
         return Type.primitive(PrimitiveType.valueOf(primitiveType.name()));
     }
 
-    static Type parseExternalType(ExternalTypeDefinition externalType, String conjurePackage, String typeName) {
+    static Type parseExternalType(
+            ExternalTypeDefinition externalType,
+            ReferenceTypeResolver nameResolver,
+            String conjurePackage,
+            String typeName) {
         if (isBearerToken(externalType) && externalType.safety().isPresent()) {
             throw new ConjureRuntimeException(
                     "Cannot mark safety of external import with base type BearerToken. BearerToken assumes Do Not"
                             + " Log.");
         }
+
         return Type.external(ExternalReference.builder()
                 .externalReference(TypeName.of(typeName, conjurePackage))
-                .fallback(ConjureParserUtils.parsePrimitiveType(externalType.baseType()))
+                .fallback(externalType.baseType().visit(new ConjureTypeParserVisitor(nameResolver)))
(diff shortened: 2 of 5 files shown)


carterkozak:
> According to the [definition spec for ExternalTypes](https://github.com/palantir/conjure/blob/master/docs/spec/conjure_definitions.md#externaltypedefinition), it is a valid definition to allow for any [ConjureType which includes ContainerType](https://github.com/palantir/conjure/blob/master/docs/spec/conjure_definitions.md#conjuretype).

The [documentation further constrains](https://github.com/palantir/conjure/blob/master/docs/spec/conjure_definitions.md#externaltypedefinition) the bounds of `ConjureType` used for external type imports:

> If the imported type is a non-primitive JSON object, then a `base-type` of any should be used.

bmarcaur:
I feel as though my decomp above is valid, but flawed due to a small assumption. I assumed you would want to keep external imports as a enduring feature. Right now they are a gap filler (borderline long term) feature. Changing this type is correct, but it also harden/expands the reasons why you would want to use this forever. Additionally, it gives so much flexibility that it lowers the incentive to do the hard work to make all of Conjure better. As such, I am going to close this PR for now.

PR #1664, opened 2024-10-21
wip: Add endpoint errors

mpritham:
## After this PR

==COMMIT_MSG==
==COMMIT_MSG==

### DIFF
--- a/conjure-api/src/main/conjure/conjure-api.yml
+++ b/conjure-api/src/main/conjure/conjure-api.yml
@@ -57,6 +57,11 @@ types:
           - CUSTOM_SERVER
       ErrorDefinition:
         fields:
+          # TODO(pm): Should there be a message here? The docs aren't the same thing: devs can add more context to them,
+          #  like remediation steps, which wouldn't be appropriate in a message. Concretely, without a message each
+          #  client needs to decide how it'll display the error to the user - could lead to inconsistency in what the
+          #  user sees. Alternatively, this is intentional and the client *should* decide how to display the error?
+          # message: Message [alias of string]
           errorName: TypeName
           docs: optional<Documentation>
           namespace: ErrorNamespace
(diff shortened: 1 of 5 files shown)


conjure-api/src/main/conjure/conjure-api.yml
```diff
@@ -57,6 +57,11 @@ types:
           - CUSTOM_SERVER
       ErrorDefinition:
         fields:
+          # TODO(pm): Should there be a message here? The docs aren't the same thing: devs can add more context to them,
```
carterkozak, line 60: Conjure errors don't provide messages by design, the name is meant to be a key such that a FE component can build a human-readable message. Providing a user-facing message would require our servers to implement internationalization/translation, which is something our current architecture is designed to handle exclusively on the frontend.
While the server can describe what failed, the FE is better positioned to contextualize the failure in order to provide good user experience.

conjure-api/src/main/conjure/conjure-api.yml
```diff
@@ -172,7 +177,18 @@ types:
...
+          # responses:
+          #   union:
+          #     - success: optional<Type>
+          #     - errors: set<TypeName>
           returns: optional<Type>
+          # TODO(pm): put up commit where ErrorName is alias for TypeName.
+          # TODO(pm): Maybe this should actually be `TypeName | ErrorDefinition` to allow devs to create new errors
+          #  inline. Think about if we want to disallow this behavior.
+          #  Let's not allow error types to be imported from other languages. Only Conjure-imports allowed.
+          errors: set<TypeName>
```
mpritham, line 191: Do we want to associate each of the `errors` with some documentation? We do this with `returns` already.

conjure-api/src/main/conjure/conjure-api.yml
mpritham: Hadn't thought about i18n. Makes sense.

mpritham:
Closing this in favor of another PR introducing similar changes.

PR #1670, opened 2024-10-29
Support specifying errors on endpoints

mpritham:
## Before this PR
Expand the Conjure specification to support specifying errors that can be thrown on each endpoint.

### Outline of work
This work is a revival of the ideas outlined in [RFC-007](https://github.com/palantir/conjure/blob/master/docs/rfc/007-endpoint-errors.md) and #816.

#### Current Behavior

* Conjure does not associate errors with endpoints. Clients are expected to keep track of errors thrown. Clients can check if the error thrown is an error they expect.
* Conjure-Java serializes the error parameters as a `Map<String, String>` where the key is the parameter name, and the value is the toString representation of the parameter [(source)](https://github.com/palantir/conjure-java-runtime-api/blob/2.55.0/errors/src/main/java/com/palantir/conjure/java/api/errors/SerializableError.java#L126).
    * This conflicts with the Conjure wire spec which says parameters will be serialized a JSON map - the parameter values should be serialized as JSON objects

#### Desired Behavior

* Conjure associates errors to endpoints, so clients know which errors they can expect when making a call to an endpoint. The addition or removal of an error thrown by an endpoint is considered an API break.
* Conjure-Java serializes the error parameters as JSON maps as per the spec.

Note: this work does not cover the generation of Dialogue client-side error handling logic.

The following PRs are relevant:
- This PR introducing the `errors` field on endpoint definitions
- [This PR in `conjure-java`](https://github.com/palantir/conjure-java/pull/2401) generating checked exception code for each endpoint error, and updating the Undertow service interfaces to throw the generated checked exceptions
- [This PR in `conjure-java-runtime-api`](https://github.com/palantir/conjure-java-runtime-api/pull/1262) introducing the abstract `CheckedServiceException` type that is used in the above `conjure-java` PR

## After this PR

==COMMIT_MSG==
Support specifying errors on endpoints
==COMMIT_MSG==

### DIFF
--- a/conjure-api/src/main/conjure/conjure-api.yml
+++ b/conjure-api/src/main/conjure/conjure-api.yml
@@ -173,6 +173,7 @@ types:
           auth: optional<AuthType>
           args: list<ArgumentDefinition>
           returns: optional<Type>
+          errors: list<EndpointError>
           docs: optional<Documentation>
           deprecated: optional<Documentation>
           markers: list<Type>
@@ -180,6 +181,10 @@ types:
       EndpointName:
         alias: string
         docs: Should be in lowerCamelCase.
+      EndpointError:
+        fields:
+          errorName: TypeName
+          docs: optional<Documentation>
       HttpMethod:
         values:
           - GET
--- a/conjure-api/src/main/conjure/conjure-api.yml
+++ b/conjure-api/src/main/conjure/conjure-api.yml
@@ -183,7 +183,7 @@ types:
         docs: Should be in lowerCamelCase.
       EndpointError:
         fields:
-          errorName: TypeName
+          error: TypeName
           docs: optional<Documentation>
       HttpMethod:
         values:
--- a/conjure-api/src/main/conjure/conjure-api.yml
+++ b/conjure-api/src/main/conjure/conjure-api.yml
@@ -183,7 +183,12 @@ types:
         docs: Should be in lowerCamelCase.
       EndpointError:
         fields:
-          error: TypeName
+          name:
+            type: string
+            docs: A reference to an error definition.
+          package:
+            type: string
+            docs: A period-delimited string of package names.
           namespace: ErrorNamespace
           docs: optional<Documentation>
       HttpMethod:
--- a/conjure-api/src/main/conjure/conjure-api.yml
+++ b/conjure-api/src/main/conjure/conjure-api.yml
@@ -184,6 +184,7 @@ types:
       EndpointError:
         fields:
           error: TypeName
+          namespace: ErrorNamespace
           docs: optional<Documentation>
       HttpMethod:
         values:
--- a/conjure-api/src/main/conjure/conjure-api.yml
+++ b/conjure-api/src/main/conjure/conjure-api.yml
@@ -181,7 +181,7 @@ types:
       EndpointName:
         alias: string
         docs: Should be in lowerCamelCase.
-      EndpointError:
+      ErrorTypeName:
         fields:
           name:
             type: string
@@ -190,6 +190,9 @@ types:
             type: string
             docs: A period-delimited string of package names.
           namespace: ErrorNamespace
+      EndpointError:
+        fields:
+          error: ErrorTypeName
           docs: optional<Documentation>
       HttpMethod:
         values:
--- a/conjure-core/src/main/java/com/palantir/conjure/defs/EndpointErrorResolver.java
+++ b/conjure-core/src/main/java/com/palantir/conjure/defs/EndpointErrorResolver.java
@@ -17,7 +17,6 @@
 package com.palantir.conjure.defs;
 
 import com.google.common.base.Preconditions;
-import com.palantir.conjure.defs.EndpointErrorResolver.ErrorResolutionResult;
 import com.palantir.conjure.parser.AnnotatedConjureSourceFile;
 import com.palantir.conjure.parser.types.ConjureType;
 import com.palantir.conjure.parser.types.ConjureTypeVisitor;
@@ -35,6 +34,7 @@ import com.palantir.conjure.parser.types.names.TypeName;
 import com.palantir.conjure.parser.types.primitive.PrimitiveType;
 import com.palantir.conjure.parser.types.reference.ForeignReferenceType;
 import com.palantir.conjure.parser.types.reference.LocalReferenceType;
+import com.palantir.conjure.spec.ErrorTypeName;
 import com.palantir.logsafe.SafeArg;
 import com.palantir.logsafe.exceptions.SafeIllegalArgumentException;
 import java.util.Map;
@@ -42,10 +42,10 @@ import java.util.Optional;
 
 /**
  * Endpoint error definitions are either references to errors defined in the same file as the endpoint definition, or in
- * imported Conjure files. This class creates a {@link com.palantir.conjure.spec.TypeName} object from a
+ * imported Conjure files. This class creates a {@link com.palantir.conjure.spec.ErrorTypeName} object from a
  * {@link com.palantir.conjure.parser.types.reference.ReferenceType} to a Conjure defined error definition.
  */
-final class EndpointErrorResolver implements ConjureTypeVisitor<ErrorResolutionResult> {
+final class EndpointErrorResolver implements ConjureTypeVisitor<ErrorTypeName> {
     private static final String UNSUPPORTED_TYPE_MESSAGE =
             "Unsupported endpoint error type. Endpoint errors must be references to a Conjure-defined error type";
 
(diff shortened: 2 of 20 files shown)


conjure-core/src/main/java/com/palantir/conjure/defs/EndpointErrorResolver.java
```diff
@@ -0,0 +1,123 @@
...
+        this.parsedErrors = parsedErrors;
+        this.defaultConjurePackage = defaultConjurePackage;
+    }
+
+    com.palantir.conjure.spec.TypeName resolve(ConjureType conjureType) {
+        Optional<TypeName> maybeTypeName = conjureType.visit(GetReferenceTypeNameVisitor.INSTANCE);
+        Preconditions.checkArgument(
+                maybeTypeName.isPresent(),
+                "Unsupported endpoint error type. Endpoint errors must be references to a Conjure-defined"
+                        + " error type");
```
carterkozak, line 57: if we throw the exception from `GetReferenceTypeNameVisitor` , we can provide more precise information. Thoughts on updating from `Optional<TypeName>` to `TypeName` and throwing in the cases that currently return `Optional.empty()`?

conjure-api/src/main/conjure/conjure-api.yml
```diff
@@ -173,13 +173,18 @@ types:
...
           docs: optional<Documentation>
           deprecated: optional<Documentation>
           markers: list<Type>
           tags: set<string>
       EndpointName:
         alias: string
         docs: Should be in lowerCamelCase.
+      EndpointError:
+        fields:
+          errorName: TypeName
```
carterkozak, line 186: If we match the pattern of `FieldDefinition`, I think perhaps this should be `error` rather than `errorName`, as it's effectively a reference to the error

conjure-core/src/main/java/com/palantir/conjure/parser/services/EndpointError.java
```diff
@@ -0,0 +1,62 @@
...
+import com.palantir.parsec.ParseException;
+import java.io.IOException;
+import java.util.Optional;
+import org.immutables.value.Value;
+
+@JsonDeserialize(using = EndpointErrorDeserializer.class)
+@Value.Immutable
+@ConjureImmutablesStyle
+public interface EndpointError {
+    @JsonProperty("error-name")
```
carterkozak, line 36: if we make the `errorName`->`error` change suggested above, we should also call this `error`

carterkozak: We should update the json-schema to support the new fields as well: https://github.com/palantir/conjure/blob/master/conjure.schema.json

carterkozak: I think there's a test that we should be able to make fail until that's updated

conjure-core/src/main/java/com/palantir/conjure/defs/EndpointErrorResolver.java
mpritham: Yeah that makes sense. The version you reviewed actually did not handle imported errors correctly. I've updated the logic to handle them correctly, and to expose the `type` in the IllegalArgumentException thrown when it's not one of the reference types.

conjure-api/src/main/conjure/conjure-api.yml
mpritham: Makes sense: it is a reference to an error defined elsewhere.

conjure-core/src/main/java/com/palantir/conjure/parser/services/EndpointError.java
mpritham: There was a failing test (`sort_types_services_and_errors`), but I don't think it references the conjure.schema.json file. I was able to recreate the `normalized.conjure.json` file and make the test pass, even though I didn't yet update the `conjure.schema.json` file.

I'll checkout how to validate a conjure definition against the JSON schema in a test.

mpritham: Just needed to add an endpoint error to [this file](https://github.com/palantir/conjure/blob/master/conjure/src/test/resources/test-service.yml) to make it fail validation.

mpritham:
Back to draft, realized I didn't update the README.

conjure-core/src/main/java/com/palantir/conjure/parser/services/EndpointError.java
```diff
@@ -0,0 +1,60 @@
...
+public interface EndpointError {
+    ConjureType error();
+
+    Optional<String> docs();
+
+    static EndpointError of(ConjureType errorName) {
+        return ImmutableEndpointError.builder().error(errorName).build();
+    }
+
+    // TODO(pm): see if there's something we can do with @JsonCreator.
```
mpritham, line 43: flagging for myself.

conjure-api/src/main/conjure/conjure-api.yml
```diff
@@ -173,13 +173,18 @@ types:
           auth: optional<AuthType>
           args: list<ArgumentDefinition>
           returns: optional<Type>
+          errors: list<EndpointError>
```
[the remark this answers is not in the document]
carterkozak: List makes sense because the same error could be listed twice in a set with slightly different docs, so we’ll have to validate uniqueness ourselves anyhow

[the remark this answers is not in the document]
mpritham: Oh true. I'll add that validation.

conjure-core/src/main/java/com/palantir/conjure/parser/services/EndpointError.java
carterkozak: I think a pair of Mode.DELEGATING static factory methods would do it, one takes a string, the other takes ImmutableEndpointError. Jackson should choose based on the shape of input.

conjure.schema.json
```diff
@@ -503,6 +516,18 @@
...
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "error": {
+          "$ref": "#/definitions/TypeName"
+        },
+        "docs": {
+          "$ref": "#/definitions/DocString"
+        }
+      }
```
carterkozak, line 529: I think we want to describe the `error` field as required when this format is used (see FieldDefinition as an example)
```suggestion
      },
      "required": [
         "error"
      ]
```

conjure-core/src/main/java/com/palantir/conjure/parser/services/EndpointError.java
mpritham: That works, thanks!

conjure.schema.json
mpritham: Makes sense, updated.

carterkozak (approved):
Let's cut an rc, nice work!

conjure-api/src/main/conjure/conjure-api.yml
```diff
@@ -173,13 +173,19 @@ types:
...
           docs: optional<Documentation>
           deprecated: optional<Documentation>
           markers: list<Type>
           tags: set<string>
       EndpointName:
         alias: string
         docs: Should be in lowerCamelCase.
+      EndpointError:
+        fields:
+          error: TypeName
```
carterkozak, line 186: I think TypeName may be a bit misleading here, there's a fair bit of overlap, but I think perhaps we should flatten the `TypeName` components (package + name) into this type, and name it `ErrorTypeName`

conjure-core/src/main/java/com/palantir/conjure/defs/EndpointErrorResolver.java
```diff
@@ -0,0 +1,158 @@
...
+            Map<Namespace, String> importProviders,
+            Map<String, AnnotatedConjureSourceFile> externalTypes) {
+        this.parsedErrors = parsedErrors;
+        this.defaultConjurePackage = defaultConjurePackage;
+        this.importProviders = importProviders;
+        this.externalTypes = externalTypes;
+    }
+
+    record ErrorResolutionResult(
+            String errorName, String package_, com.palantir.conjure.spec.ErrorNamespace namespace) {}
```
carterkozak, line 69: Would it make sense to use the new ErrorTypeName type here?

conjure-api/src/main/conjure/conjure-api.yml
```diff
@@ -173,13 +173,24 @@ types:
...
           returns: optional<Type>
+          errors: list<EndpointError>
           docs: optional<Documentation>
           deprecated: optional<Documentation>
           markers: list<Type>
           tags: set<string>
       EndpointName:
         alias: string
         docs: Should be in lowerCamelCase.
+      EndpointError:
```
carterkozak, line 184: Can we rename this to `ErrorTypeName`? It's basically equivalent to `TypeName`, but for errors

carterkozak: sorry: this type makes sense, but should encapsulate a pair of ErrorTypeName (package/namespace/name) + docs

conjure-core/src/main/java/com/palantir/conjure/defs/EndpointErrorResolver.java
mpritham: Yep. Updated.

conjure-api/src/main/conjure/conjure-api.yml
mpritham: Updated!

carterkozak (dismissed):
Thanks!

carterkozak (dismissed):
Great work!

carterkozak (approved):

PR #1685, opened 2024-11-18
Enforce that Conjure error definitions are unique

carterkozak:
## Before this PR
No validation that the same error name wasn't used in multiple definitions. Over the wire, clients cannot distinguish between multiple errors when the namespace and name (excluding package) match.

==COMMIT_MSG==
Enforce that Conjure error definitions are unique
==COMMIT_MSG==

## Possible downsides?
This could prevent some repos from rolling out changes if they already have duplicate errors. However this is fairly unlikely, and the failure is self-documenting.

### DIFF
--- a/conjure-core/src/main/java/com/palantir/conjure/defs/validator/ConjureDefinitionValidator.java
+++ b/conjure-core/src/main/java/com/palantir/conjure/defs/validator/ConjureDefinitionValidator.java
@@ -18,9 +18,13 @@ package com.palantir.conjure.defs.validator;
 
 import com.google.common.base.Joiner;
 import com.google.common.base.Preconditions;
+import com.google.common.collect.ArrayListMultimap;
 import com.google.common.collect.HashMultimap;
+import com.google.common.collect.ListMultimap;
 import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
 import com.google.common.collect.Multimap;
+import com.google.common.collect.Multimaps;
 import com.palantir.conjure.defs.Conjure;
 import com.palantir.conjure.defs.SafetyDeclarationRequirements;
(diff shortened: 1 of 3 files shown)


conjure-core/src/main/java/com/palantir/conjure/defs/validator/ConjureDefinitionValidator.java
```diff
@@ -505,4 +511,45 @@ private static boolean isSafetyMarker(Type type) {
...
+            List<String> packages = definitions.stream()
+                    .map(errorDef -> errorDef.getErrorName().getPackage())
+                    .distinct()
+                    .toList();
+            if (definitions.size() == packages.size()) {
+                return packages.stream().collect(Collectors.joining(", ", "in packages: ", ""));
+            }
+            return definitions.stream()
+                    .map(ErrorDefinition::toString)
+                    .collect(Collectors.joining("\n\t", "defined %n times:\n\t", ""));
```
mpritham, line 552: Do we need to pass in the `definitions.size()` somewhere so `%n` is printed as expected? Or does `Collectors.joining` do something special internally

mpritham: hm, I don't see anything special in the docs.

```suggestion
                    .collect(Collectors.joining("\n\t", String.format("defined %d times:\n\t", definitions.size()), ""));
```

carterkozak: err, I'm just dumb. fixed!

mpritham (approved):
Nit comment.

Approving to unblock. Thanks!

PR #1715, opened 2025-01-11
Precompute valid regexes String in FieldNameValidator.validate()

ash211:
## Before this PR
This String containing the list of valid case pattern regexes is recomputed for every call to FieldNameValidator.validate(), and appears in JFRs of the gradle daemon:

Screenshot 2025-01-10 at 10 07 53 PM

(not a lot, but a little at least)

## After this PR
==COMMIT_MSG==
Precompute valid regexes String in FieldNameValidator.validate()
==COMMIT_MSG==

## Possible downsides?
None known

### DIFF
--- a/conjure-core/src/main/java/com/palantir/conjure/defs/validator/FieldNameValidator.java
+++ b/conjure-core/src/main/java/com/palantir/conjure/defs/validator/FieldNameValidator.java
@@ -18,6 +18,7 @@ package com.palantir.conjure.defs.validator;
 
 import com.google.common.base.Preconditions;
 import com.palantir.conjure.CaseConverter;
+import com.palantir.conjure.CaseConverter.Case;
 import com.palantir.conjure.spec.FieldName;
 import java.util.Arrays;
 import org.apache.commons.lang3.StringUtils;
@@ -27,6 +28,8 @@ import org.slf4j.LoggerFactory;
 public final class FieldNameValidator {
 
     private static final Logger log = LoggerFactory.getLogger(FieldNameValidator.class);
+    // Precompute this String because otherwise it is computed for every call to validate() and appears in JFRs.
+    private static final String CASE_VALUES = Arrays.toString(Case.values());
 
     private FieldNameValidator() {}
 
@@ -53,7 +56,7 @@ public final class FieldNameValidator {
                         || CaseConverter.SNAKE_CASE_PATTERN.matches(fieldName.get()),
                 "FieldName \"%s\" must follow one of the following patterns: %s",
                 fieldName,
-                Arrays.toString(CaseConverter.Case.values()));
+                CASE_VALUES);
 
         if (!matchesCamelCase) {
             log.warn(
--- a/conjure-core/src/test/java/com/palantir/conjure/defs/validator/FieldNameValidatorTest.java
+++ b/conjure-core/src/test/java/com/palantir/conjure/defs/validator/FieldNameValidatorTest.java
@@ -40,7 +40,7 @@ public final class FieldNameValidatorTest {
     }
 
     @Test
-    public void testInvalidNames() throws Exception {
+    public void testInvalidNames() {
         for (String invalid : new String[] {
             "UpperCamelCase",
             "Upper-Kebab-Case",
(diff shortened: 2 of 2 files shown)


conjure-core/src/test/java/com/palantir/conjure/defs/validator/FieldNameValidatorTest.java
```diff
@@ -55,18 +55,22 @@ public void testInvalidNames() throws Exception {
...
             assertThatThrownBy(() -> FieldNameValidator.validate(FieldName.of(invalid)))
                     .isInstanceOf(IllegalArgumentException.class)
-                    .hasMessageContaining(
-                            String.format("FieldName \"%s\" must follow one of the following patterns", invalid));
+                    .hasMessageContaining(String.format(
+                            "FieldName \"%s\" must follow one of the following patterns: "
+                                    + "[LOWER_CAMEL_CASE[^[a-z]([A-Z]{1,2}[a-z0-9]|[a-z0-9])*[A-Z]?$], "
+                                    + "KEBAB_CASE[^[a-z]((-[a-z]){1,2}[a-z0-9]|[a-z0-9])*(-[a-z])?$], "
+                                    + "SNAKE_CASE[^[a-z]((_[a-z]){1,2}[a-z0-9]|[a-z0-9])*(_[a-z])?$]]",
+                            invalid));
```
pkoenig10, line 63: Why did we change this? It may create unnecessary work if we later update these patterns. We don't particularly care to test this part of the message exactly. Although I'm generally in favor of validating full messages.

Did you intend to change `hasMessageContaining` to `hasMessage`?

pkoenig10 (approved):

conjure-core/src/test/java/com/palantir/conjure/defs/validator/FieldNameValidatorTest.java
ash211: Ah good point.  I'd wanted to confirm I hadn't changed / broken this part of the message, but pasting the full contents makes it a [change detector test](https://testing.googleblog.com/2015/01/testing-on-toilet-change-detector-tests.html).  

Instead I've now rewritten the assertions to assert it contains `LOWER_CAMEL_CASE` and `[A-Z]` as a measure of containing the name of the regexes plus the contents of the regexes

conjure-core/src/main/java/com/palantir/conjure/defs/validator/FieldNameValidator.java
```diff
@@ -53,7 +56,7 @@ public static void validate(FieldName fieldName) {
                         || CaseConverter.SNAKE_CASE_PATTERN.matches(fieldName.get()),
                 "FieldName \"%s\" must follow one of the following patterns: %s",
                 fieldName,
-                Arrays.toString(CaseConverter.Case.values()));
+                CASE_VALUES);
```
carterkozak, line 59: instead of precomputing the components of the failure message, what if we simplify this to replace the preconditions check with a simple conditional? This way the error message components are only accessed when needed, and all the relevant data lives in the same place.

In most executions, the precomputed `Arrays.toString(Case.values())` isn't read, so it's wasted computation (albeit small and executed once).

conjure-core/src/test/java/com/palantir/conjure/defs/validator/FieldNameValidatorTest.java
```diff
@@ -56,17 +56,19 @@ public void testInvalidNames() throws Exception {
             assertThatThrownBy(() -> FieldNameValidator.validate(FieldName.of(invalid)))
                     .isInstanceOf(IllegalArgumentException.class)
                     .hasMessageContaining(
-                            String.format("FieldName \"%s\" must follow one of the following patterns", invalid));
+                            String.format("FieldName \"%s\" must follow one of the following patterns: ", invalid))
+                    .hasMessageContaining("LOWER_CAMEL_CASE")
+                    .hasMessageContaining("[A-Z]");
```
carterkozak, line 61: Let's use a single  `hasMessageContainingAll` so that on failure, we get the full context:
```java
hasMessageContainingAll(
    String.format("FieldName \"%s\" must follow one of the following patterns: ", invalid),
    "LOWER_CAMEL_CASE",
    "[A-Z]");

ash211:
@carterkozak applied your suggestions -- thanks!

PR #1752, opened 2025-03-04
Fix complex query param validation

kunalrkak:
## Before this PR

Conjure endpoint validator allows nested containers in an endpoint query argument, for example: 
```
ArgumentDefinition:
    type: optional<list<string>>
    param-type: query
```
Query params cannot be nested container types, as container inner types must be serialized using PLAIN format, which only supports primitives + enums.

The example above in particular creates ambiguity in the wire format (since types with multiple elements are encoded to urls as a series of individual elements: `?arg=foo,arg=bar`, while optional types are simply omitted from the query string if the value is not present [[docs](https://palantir.github.io/conjure/#/docs/spec/wire?id=_22-query-parameters)]). In this case, an absence of the argument could be either the empty optional or a present optional with an empty list.

## After this PR

==COMMIT_MSG==
Conjure endpoint validator throws for nested container query arguments.
==COMMIT_MSG==

### DIFF
--- a/conjure-core/src/test/java/com/palantir/conjure/defs/validator/EndpointDefinitionTest.java
+++ b/conjure-core/src/test/java/com/palantir/conjure/defs/validator/EndpointDefinitionTest.java
@@ -254,6 +254,159 @@ public final class EndpointDefinitionTest {
     }
 
     @Test
+    public void testQueryParamValidatorNestedContainer_list_optional() {
+        EndpointDefinition.Builder definition = EndpointDefinition.builder()
+                .args(ImmutableList.of(ArgumentDefinition.builder()
+                        .argName(ArgumentName.of("paramName"))
+                        .type(Type.list(
+                                ListType.of(Type.optional(OptionalType.of(Type.primitive(PrimitiveType.STRING))))))
+                        .paramType(ParameterType.query(QueryParameterType.of(ParameterId.of("value"))))
+                        .build()))
+                .endpointName(ENDPOINT_NAME)
+                .httpMethod(HttpMethod.GET)
+                .httpPath(HttpPath.of("/path"));
+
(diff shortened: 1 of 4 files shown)


carterkozak:
I think the clearest existing docs which support this are from [ArgumentDefinition](https://palantir.github.io/conjure/#/docs/spec/conjure_definitions?id=argumentdefinition):

> If the param-type is query then the de-aliased type MUST be an enum or a primitive (except binary and bearertoken), or a container (list, set, optional) of one of these

conjure-core/src/test/java/com/palantir/conjure/defs/validator/EndpointDefinitionTest.java
```diff
@@ -253,6 +253,159 @@ public void testQueryParamValidatorBinaryParamInContainer_optional() {
...
+                                ListType.of(Type.optional(OptionalType.of(Type.primitive(PrimitiveType.STRING))))))
+                        .build())));
+
+        assertThatThrownBy(() -> EndpointDefinitionValidator.validateAll(definition.build(), dealiasingVisitor))
+                .isInstanceOf(IllegalArgumentException.class)
+                .hasMessage(
+                        "Query parameters must be enums or primitives when de-aliased, or containers of these (list,"
+                                + " sets, optionals): 'paramName' is not allowed on endpoint 'test{http: GET /path}'");
+    }
+
```
mpritham, line 369: Shall we add a test for `container<enum>` to exercise the case where we recurse and the de-aliasing visitor returns `Either.left`?

kunalrkak: good call, will add!

mpritham (approved):
Thanks!

PR #2109, opened 2026-06-02
Use quicktheories for property-based-testing

schlosna:
## Before this PR

Used jqwik for property based testing in one unit test. [jqwik 1.10.0 attempts to prompt inject to prevent LLM usage](https://github.com/jqwik-team/jqwik/issues/708).

See https://github.com/palantir/tritium/issues/2472

## After this PR

==COMMIT_MSG==
Use quicktheories for property based testing
==COMMIT_MSG==


conjure-generator-common/src/test/java/com/palantir/conjure/PatternTestGenerators.java
```diff
@@ -0,0 +1,45 @@
...
+
+package com.palantir.conjure;
+
+import java.util.ArrayList;
+import java.util.List;
+import org.quicktheories.core.Gen;
+import org.quicktheories.generators.Generate;
+import org.quicktheories.generators.SourceDSL;
+
+final class PatternTestGenerators {
```
pkoenig10, line 25:
```suggestion
final class Generators {
```

conjure-generator-common/src/test/java/com/palantir/conjure/PatternTestGenerators.java
```diff
@@ -0,0 +1,45 @@
...
+
+    static Gen<String> stringsFromChars(String validChars, int maxLength) {
+        List<Character> chars = new ArrayList<>(validChars.length());
+        for (int i = 0; i < validChars.length(); i++) {
+            chars.add(validChars.charAt(i));
+        }
```
pkoenig10, line 31: In https://github.com/palantir/resource-identifier/pull/772, we have identical behavior implemented as:

```java
Gen<Character> characterGen =
        Generate.pick(chars.chars().mapToObj(c -> (char) c).toList());
```

I don't have a preference for which implementation we use, but we should try to be consistent.

conjure-generator-common/src/test/java/com/palantir/conjure/PatternTestGenerators.java
```diff
@@ -0,0 +1,45 @@
...
+        return SourceDSL.lists()
+                .of(Generate.pick(chars))
+                .ofSizeBetween(0, maxLength)
+                .map(list -> {
+                    StringBuilder sb = new StringBuilder(list.size());
+                    for (Character c : list) {
+                        sb.append(c.charValue());
+                    }
+                    return sb.toString();
+                });
```
pkoenig10, line 41: This is also inconsistent with https://github.com/palantir/resource-identifier/pull/772, where the variable is named `ch` and we use `forEach` instead of a for loop.

pkoenig10 (approved):

conjure-generator-common/src/test/java/com/palantir/conjure/PatternTestGenerators.java
schlosna: https://github.com/palantir/conjure/pull/2110

PR #2149, opened 2026-07-24
Add ConjureIrValidator to validate deserialized Conjure IR

jppal9000:
## Before this PR
`ConjureDefinitionValidator.validateAll` runs only cross-cutting checks (unique
names, version, recursion, nested optionals, illegal map keys, log-safety). The
per-type identifier, field, package and HTTP-path validators otherwise run only
while parsing source, so a consumer holding an already-built (e.g. deserialized)
`ConjureDefinition` has no single entry point applying the validation the compiler
applies during parse.

## After this PR
Adds `ConjureIrValidator`, which validates a fully-built `ConjureDefinition` the way
the parser does — per-type name/format validators, field name and field definition
validators, package and HTTP-path validators, endpoint validators, then the existing
cross-cutting `ConjureDefinitionValidator.validateAll`. Consumers that ingest compiled
IR can enforce the same spec conformance without re-parsing. The parser
(`ConjureParserUtils`) is unchanged.

Companion to palantir/conjure-python#1323: per review there, spec-conformance
validation of the IR belongs here, while the generator validates the code it emits.

## Possible downsides?
None expected — it composes existing validators without modifying them or the parse
path, and is purely additive (a new class).

### DIFF
--- a/conjure-core/src/main/java/com/palantir/conjure/defs/validator/ConjureIrValidator.java
+++ b/conjure-core/src/main/java/com/palantir/conjure/defs/validator/ConjureIrValidator.java
@@ -17,6 +17,7 @@
 package com.palantir.conjure.defs.validator;
 
 import com.palantir.conjure.defs.SafetyDeclarationRequirements;
+import com.palantir.conjure.exceptions.ConjureIllegalStateException;
 import com.palantir.conjure.spec.AliasDefinition;
 import com.palantir.conjure.spec.ConjureDefinition;
 import com.palantir.conjure.spec.EnumDefinition;
@@ -42,7 +43,6 @@ public final class ConjureIrValidator {
 
     private ConjureIrValidator() {}
 
-    /** Validates using {@link SafetyDeclarationRequirements#ALLOWED}. */
     public static void validate(ConjureDefinition definition) {
         validate(definition, SafetyDeclarationRequirements.ALLOWED);
     }
@@ -54,7 +54,8 @@ public final class ConjureIrValidator {
         definition.getTypes().forEach(type -> type.accept(TypeValidator.INSTANCE));
         definition.getErrors().forEach(error -> {
             PackageValidator.validate(error.getErrorName().getPackage());
-            // Field names/definitions must be validated before ErrorDefinitionValidator, which normalizes field
+            ErrorNamespaceValidator.validate(error.getNamespace());
+            // Field names/definitions should be validated before ErrorDefinitionValidator, which normalizes field
             // names via FieldNameValidator#toCase and assumes they already conform to the FieldName grammar.
             validateFields(error.getSafeArgs());
             validateFields(error.getUnsafeArgs());
@@ -72,7 +73,6 @@ public final class ConjureIrValidator {
         ConjureDefinitionValidator.validateAll(definition, safetyRequirements);
     }
 
-    /** Validates each field's name and definition, matching the checks the parser applies via its shared helper. */
     private static void validateFields(List<FieldDefinition> fields) {
         fields.forEach(field -> {
             FieldNameValidator.validate(field.getFieldName());
@@ -102,7 +102,7 @@ public final class ConjureIrValidator {
         public Void visitObject(ObjectDefinition value) {
             PackageValidator.validate(value.getTypeName().getPackage());
             TypeNameValidator.validate(value.getTypeName());
-            // Field names/definitions must be validated before ObjectDefinitionValidator, which normalizes field
+            // Field names/definitions should be validated before ObjectDefinitionValidator, which normalizes field
             // names via FieldNameValidator#toCase and assumes they already conform to the FieldName grammar.
             validateFields(value.getFields());
             ObjectDefinitionValidator.validate(value);
@@ -119,8 +119,8 @@ public final class ConjureIrValidator {
         }
 
         @Override
-        public Void visitUnknown(String _unknownType) {
-            return null;
+        public Void visitUnknown(String unknownType) {
+            throw new ConjureIllegalStateException("Unknown type: " + unknownType);
         }
     }
 }
--- a/conjure-core/src/main/java/com/palantir/conjure/defs/validator/ConjureIrValidator.java
+++ b/conjure-core/src/main/java/com/palantir/conjure/defs/validator/ConjureIrValidator.java
@@ -1,0 +1,126 @@
+/*
+ * (c) Copyright 2018 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.conjure.defs.validator;
+
+import com.palantir.conjure.defs.SafetyDeclarationRequirements;
+import com.palantir.conjure.spec.AliasDefinition;
+import com.palantir.conjure.spec.ConjureDefinition;
+import com.palantir.conjure.spec.EnumDefinition;
+import com.palantir.conjure.spec.FieldDefinition;
+import com.palantir.conjure.spec.ObjectDefinition;
+import com.palantir.conjure.spec.TypeDefinition;
+import com.palantir.conjure.spec.UnionDefinition;
+import com.palantir.conjure.visitor.DealiasingTypeVisitor;
+import com.palantir.conjure.visitor.TypeDefinitionVisitor;
+import java.util.List;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+/**
+ * Validates an already-deserialized {@link ConjureDefinition} the way the compiler validates during parse.
+ *
+ * <p>{@link ConjureDefinitionValidator#validateAll} runs only cross-cutting checks (unique names, version,
+ * recursion, nested optionals, illegal map keys, log-safety). The per-type identifier and format validators are
+ * otherwise invoked only while parsing source. This entry point runs both, so a consumer that deserializes compiled
+ * IR (rather than parsing source) can enforce the same spec conformance.
+ */
+public final class ConjureIrValidator {
+
+    private ConjureIrValidator() {}
+
+    /** Validates using {@link SafetyDeclarationRequirements#ALLOWED}. */
+    public static void validate(ConjureDefinition definition) {
+        validate(definition, SafetyDeclarationRequirements.ALLOWED);
+    }
+
+    public static void validate(ConjureDefinition definition, SafetyDeclarationRequirements safetyRequirements) {
+        DealiasingTypeVisitor dealiasingTypeVisitor = new DealiasingTypeVisitor(definition.getTypes().stream()
+                .collect(Collectors.toMap(type -> type.accept(TypeDefinitionVisitor.TYPE_NAME), Function.identity())));
+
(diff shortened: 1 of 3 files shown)


conjure-core/src/test/java/com/palantir/conjure/defs/validator/ConjureIrValidatorTest.java
```diff
@@ -0,0 +1,240 @@
...
+                .version(1)
+                .types(List.of(TypeDefinition.enum_(EnumDefinition.builder()
+                        .typeName(ENUM_NAME)
+                        .values(List.of(
+                                EnumValueDefinition.builder().value("RED").build(),
+                                EnumValueDefinition.builder().value("not valid").build()))
+                        .build())))
+                .build();
+
+        assertThatThrownBy(() -> ConjureIrValidator.validate(definition)).isInstanceOf(IllegalArgumentException.class);
```
dtobin: minor: add `.hasMessage` checks to these tests to verify that the failure it's hitting is the one that we intend to be testing.

conjure-core/src/test/java/com/palantir/conjure/defs/validator/ConjureIrValidatorTest.java
```diff
@@ -0,0 +1,240 @@
...
+                        .values(List.of(
+                                EnumValueDefinition.builder().value("RED").build()))
+                        .build())))
+                .build();
+
+        assertThatThrownBy(() -> ConjureIrValidator.validate(definition)).isInstanceOf(IllegalArgumentException.class);
+    }
+
+    @Test
+    public void acceptsValidDefinition() {
```
dtobin: minor: this makes more sense as the first test in the file

conjure-core/src/test/java/com/palantir/conjure/defs/validator/ConjureIrValidatorTest.java
```diff
@@ -0,0 +1,240 @@
...
+                                .httpPath(HttpPath.of("widget"))
+                                .build())
+                        .build()))
+                .build();
+
+        assertThatThrownBy(() -> ConjureIrValidator.validate(definition)).isInstanceOf(IllegalArgumentException.class);
+    }
+
+    @Test
+    public void acceptsValidDefinitionWithObjectAndService() {
```
dtobin: minor: just combine the positive test cases into one test (including the types and services)

conjure-core/src/test/java/com/palantir/conjure/defs/validator/ConjureIrValidatorTest.java
```diff
@@ -0,0 +1,240 @@
...
+                                .type(Type.primitive(PrimitiveType.STRING))
+                                .build())
+                        .build())))
+                .build();
+
+        assertThatThrownBy(() -> ConjureIrValidator.validate(definition)).isInstanceOf(IllegalArgumentException.class);
+    }
+
+    @Test
+    public void rejectsInvalidHttpPath() {
```
dtobin: minor: I'd organize these tests in the same order as the logic in `ConjureIrValidator.validate` - so types, then errors, then services, then ConjureDefinitionValidator.validateAll.

conjure-core/src/test/java/com/palantir/conjure/defs/validator/ConjureIrValidatorTest.java
```diff
@@ -0,0 +1,240 @@
...
+import com.palantir.conjure.spec.QueryParameterType;
+import com.palantir.conjure.spec.ServiceDefinition;
+import com.palantir.conjure.spec.Type;
+import com.palantir.conjure.spec.TypeDefinition;
+import com.palantir.conjure.spec.TypeName;
+import com.palantir.conjure.spec.UnionDefinition;
+import java.util.List;
+import org.junit.jupiter.api.Test;
+
+public final class ConjureIrValidatorTest {
```
dtobin, line 50: minor: add at some test cases covering the `ConjureDefinitionValidator.validateAll` codepaths

conjure-core/src/main/java/com/palantir/conjure/defs/validator/ConjureIrValidator.java
```diff
@@ -0,0 +1,126 @@
...
+ * <p>{@link ConjureDefinitionValidator#validateAll} runs only cross-cutting checks (unique names, version,
+ * recursion, nested optionals, illegal map keys, log-safety). The per-type identifier and format validators are
+ * otherwise invoked only while parsing source. This entry point runs both, so a consumer that deserializes compiled
+ * IR (rather than parsing source) can enforce the same spec conformance.
+ */
+public final class ConjureIrValidator {
+
+    private ConjureIrValidator() {}
+
+    /** Validates using {@link SafetyDeclarationRequirements#ALLOWED}. */
```
dtobin: minor: this comment is not necessary

conjure-core/src/main/java/com/palantir/conjure/defs/validator/ConjureIrValidator.java
```diff
@@ -0,0 +1,126 @@
...
+            PackageValidator.validate(value.getTypeName().getPackage());
+            TypeNameValidator.validate(value.getTypeName());
+            validateFields(value.getUnion());
+            UnionDefinitionValidator.validateAll(value);
+            return null;
+        }
+
+        @Override
+        public Void visitUnknown(String _unknownType) {
+            return null;
```
dtobin: Throw here

conjure-core/src/main/java/com/palantir/conjure/defs/validator/ConjureIrValidator.java
```diff
@@ -0,0 +1,126 @@
...
+        validate(definition, SafetyDeclarationRequirements.ALLOWED);
+    }
+
+    public static void validate(ConjureDefinition definition, SafetyDeclarationRequirements safetyRequirements) {
+        DealiasingTypeVisitor dealiasingTypeVisitor = new DealiasingTypeVisitor(definition.getTypes().stream()
+                .collect(Collectors.toMap(type -> type.accept(TypeDefinitionVisitor.TYPE_NAME), Function.identity())));
+
+        definition.getTypes().forEach(type -> type.accept(TypeValidator.INSTANCE));
+        definition.getErrors().forEach(error -> {
+            PackageValidator.validate(error.getErrorName().getPackage());
```
dtobin, line 56: Also call `ErrorNamespaceValidator`

conjure-core/src/main/java/com/palantir/conjure/defs/validator/ConjureIrValidator.java
```diff
@@ -0,0 +1,126 @@
...
+            // Field names/definitions must be validated before ErrorDefinitionValidator, which normalizes field
```
dtobin:
```suggestion
            // Field names/definitions should be validated before ErrorDefinitionValidator, which normalizes field
```

conjure-core/src/main/java/com/palantir/conjure/defs/validator/ConjureIrValidator.java
```diff
@@ -0,0 +1,126 @@
...
+            service.getEndpoints().forEach(endpoint -> {
+                HttpPathValidator.validate(endpoint.getHttpPath());
+                EndpointDefinitionValidator.validateAll(endpoint, dealiasingTypeVisitor);
+            });
+        });
+
+        ConjureDefinitionValidator.validateAll(definition, safetyRequirements);
+    }
+
+    /** Validates each field's name and definition, matching the checks the parser applies via its shared helper. */
```
dtobin: minor: this comment is not necessary

conjure-core/src/main/java/com/palantir/conjure/defs/validator/ConjureIrValidator.java
```diff
@@ -0,0 +1,126 @@
...
+            TypeNameValidator.validate(value.getTypeName());
+            EnumDefinitionValidator.validateAll(value);
+            return null;
+        }
+
+        @Override
+        public Void visitObject(ObjectDefinition value) {
+            PackageValidator.validate(value.getTypeName().getPackage());
+            TypeNameValidator.validate(value.getTypeName());
+            // Field names/definitions must be validated before ObjectDefinitionValidator, which normalizes field
```
dtobin:
```suggestion
            // Field names/definitions should be validated before ObjectDefinitionValidator, which normalizes field
```

conjure-core/src/test/java/com/palantir/conjure/defs/validator/ConjureIrValidatorTest.java
jppal9000: Done — each rejection test now asserts the specific failure via `.hasMessageContaining(...)`, so a test only passes on the intended validation failure.

conjure-core/src/test/java/com/palantir/conjure/defs/validator/ConjureIrValidatorTest.java
jppal9000: Done — the combined positive case (`acceptsValidDefinition`) is now the first test in the file.

conjure-core/src/test/java/com/palantir/conjure/defs/validator/ConjureIrValidatorTest.java
jppal9000: Done — folded into a single `acceptsValidDefinition` covering an enum, object, and union type plus an error and a service, asserting both `validate` overloads don't throw.

conjure-core/src/test/java/com/palantir/conjure/defs/validator/ConjureIrValidatorTest.java
jppal9000: Done — reordered to match `validate()`: types, then errors, then services, then `ConjureDefinitionValidator.validateAll`, with section comments for each group.

conjure-core/src/test/java/com/palantir/conjure/defs/validator/ConjureIrValidatorTest.java
jppal9000: Done — added `rejectsUnsupportedVersion` (version check) and `rejectsDuplicateServiceNames` (unique service names), both failing inside `ConjureDefinitionValidator.validateAll`.

conjure-core/src/main/java/com/palantir/conjure/defs/validator/ConjureIrValidator.java
jppal9000: Done — removed.

conjure-core/src/main/java/com/palantir/conjure/defs/validator/ConjureIrValidator.java
jppal9000: Done — `visitUnknown` now throws `ConjureIllegalStateException`. I left it without a dedicated unit test since the generated `TypeDefinition` builder only produces the known variants, so there's no supported way to construct an unknown-variant instance to exercise this branch.

conjure-core/src/main/java/com/palantir/conjure/defs/validator/ConjureIrValidator.java
jppal9000: Done — added `ErrorNamespaceValidator.validate(error.getNamespace())`, with coverage in `rejectsInvalidErrorNamespace`.

conjure-core/src/main/java/com/palantir/conjure/defs/validator/ConjureIrValidator.java
jppal9000: Applied, thanks.

conjure-core/src/main/java/com/palantir/conjure/defs/validator/ConjureIrValidator.java
jppal9000: Done — removed.

conjure-core/src/main/java/com/palantir/conjure/defs/validator/ConjureIrValidator.java
jppal9000: Applied, thanks.

PR #2152, opened 2026-07-28
Accept constructs the compiler allows in conjure.schema.json

ItsMacto:
## Before this PR

`conjure.schema.json` is published to SchemaStore and used by editors to validate Conjure definitions as
they are written. It is not read by the compiler, so it can drift from the parser without anything
failing, and it has: six constructs that `conjure compile` accepts are reported as errors, so editors
underline valid definitions.

| Construct | Why it is legal |
| --- | --- |
| namespace alias containing a digit, e.g. `v2.ImportedObject` | `Namespace` is `^[_a-zA-Z][_a-zA-Z0-9]*$`, documented at `docs/spec/conjure_definitions.md` line 58 and already used by this schema's own `conjure-imports` key pattern |
| `safety` on an external import | `ExternalTypeDefinition#safety()` is `Optional<LogSafetyDefinition>`, and the value reaches the IR as `external.safety` |
| `markers` on an endpoint | `EndpointDefinition#markers()` is `Set<ConjureType>` |
| `docs:` with no value | `Optional<String> docs()`, so `null` is an absent key |
| `tags:` with no value | `Set<String> tags()` |
| argument `markers:` with no value | `Set<ConjureType> markers()` |

The last three are what you get when a key is typed ahead of its value, a normal intermediate state while
editing.

The fixtures added here are the reproductions: each fails against the current schema and compiles cleanly
with `conjure compile`.

## After this PR

==COMMIT_MSG==
Allow the JSON schema to accept six constructs the Conjure compiler already accepts: namespaced type
references whose namespace contains a digit or underscore, `safety` on an external import, `markers` on
an endpoint, and `docs`/`tags`/`markers` keys written with no value.
==COMMIT_MSG==

| Node | Change |
| --- | --- |
| `TypeName.pattern` | namespace prefix `([a-zA-Z]+[.])?` to `([_a-zA-Z][_a-zA-Z0-9]*[.])?` |
| `ExternalTypeDefinition` | add `safety: {$ref: LogSafety}` |
| `EndpointDefinition` | add `markers`, items `ConjureType`, nullable |
| `EndpointDefinition.tags` | `type: array` to `type: [array, null]` |
| `ArgumentDefinition.tags`, `ArgumentDefinition.markers` | `type: array` to `type: [array, null]` |
| `DocString` | `type: string` to `type: [string, null]` |

The PascalCase portion of `TypeName` is untouched, so `HTTPConfig` is still rejected, matching
`TypeName`'s own `^[A-Z][a-z0-9]+([A-Z][a-z0-9]+)*$`.

Seven fixtures are added under `conjure/src/test/resources/json-schema/`, picked up by the recursive
`Files.walk` in `ConjureJsonSchemaTest`. Two are negative fixtures asserting the loosened patterns did not
go too far: `HTTPConfig` is still rejected as a type name, and `v-5.Foo` and `1v.Foo` are still rejected
as type references.

## Possible downsides?

The schema is not an input to the compiler, so this cannot change the outcome of a build. The only risk is
a schema that accepts something it should reject, and both loosened patterns are bounded by the negative
fixtures above.

Two things worth flagging rather than leaving to be found:

1. **Endpoint `markers` is supported by the parser but is not in the spec.** The `EndpointDefinition`
   table in `docs/spec/conjure_definitions.md` lists `http`, `auth`, `returns`, `errors`, `args`, `docs`,
   `deprecated` and `tags`, but not `markers`, even though `EndpointDefinition#markers()` exists and the
   value reaches the IR. This PR follows the parser. Happy to send a follow-up adding the row if you would
   rather the spec led.

2. **`safety` on an external import has the same gap.** The `ExternalTypeDefinition` table lists only
   `base-type` and `external`. Same offer.

### DIFF
--- a/conjure/src/test/resources/json-schema/null-values.yml
+++ b/conjure/src/test/resources/json-schema/null-values.yml
@@ -1,33 +1,29 @@
 # Keys written with no value parse as null. The compiler treats these the same as
-# an absent key, so the schema must accept them.
+# an absent key, but the schema is deliberately stricter: an empty key is not
+# something an author means to keep, so editors should flag it.
 types:
   definitions:
     default-package: test.api
     objects:
       NullDocsObject:
+        # SCHEMA(type): an empty docs key is not valid
         docs:
         fields:
-          value:
-            type: string
-            docs:
+          value: string
 services:
   NullValueService:
     name: Null Value Service
     package: test.api
     base-path: /null
+    # SCHEMA(type): an empty docs key is not valid
     docs:
     endpoints:
       getValue:
         http: GET /value
         returns: NullDocsObject
+        # SCHEMA(type): an empty docs key is not valid
         docs:
+        # SCHEMA(type): an empty tags key is not valid
         tags:
+        # SCHEMA(type): an empty markers key is not valid
         markers:
-        args:
-          filter:
-            type: string
-            param-type: query
-            param-id: filter
-            docs:
-            tags:
-            markers:
--- a/conjure/src/test/resources/json-schema/null-values.yml
+++ b/conjure/src/test/resources/json-schema/null-values.yml
@@ -1,0 +1,33 @@
+# Keys written with no value parse as null. The compiler treats these the same as
+# an absent key, so the schema must accept them.
+types:
+  definitions:
+    default-package: test.api
+    objects:
+      NullDocsObject:
+        docs:
+        fields:
+          value:
+            type: string
+            docs:
+services:
+  NullValueService:
+    name: Null Value Service
+    package: test.api
+    base-path: /null
+    docs:
+    endpoints:
+      getValue:
+        http: GET /value
+        returns: NullDocsObject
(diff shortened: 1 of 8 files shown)


conjure/src/test/resources/json-schema/null-values.yml
```diff
@@ -0,0 +1,33 @@
+# Keys written with no value parse as null. The compiler treats these the same as
```
mpritham, line 1: I understand the argument for 1:1 correspondence between the schema and what the compiler accepts, but given our usage of this schema only in editors, what are your thoughts on continuing to flag these null values? I don't think a dev (or LLM) would ever want to keep around an empty key.

ItsMacto: Resolved, let me know if to wordy. Made the comment to document it

mpritham (approved):

